PR #9, opened 2018-06-12
Proposed Metadata Structure

Zombiefruit:


packages/meta-data-tools/src/types.ts
```diff
@@ -1,16 +1,35 @@
...
+import { ComponentType } from 'react';
 
-export interface MetaDataToolsDefinition {
-  metaData: Map<typeof React.Component | React.StatelessComponent<any>, MetaDataDefinition>;
-  describe: (comp: typeof React.Component | React.StatelessComponent<any>) => MetaDataDefinition;
+export interface IRegistry<AssetMap = any> {
+  metadata: IMetadata;
+  describeComponent:<Props> (comp: ComponentType<Props>) => IComponentMetadata<Props>;
+  describeAsset:<AssetType extends keyof AssetMap, Asset extends AssetMap[AssetType]> (asset: Asset, type: AssetType, name: string, description?: string) => IAssetMetadata<AssetType>;
+  describeTheme: (style: any) => IStyleMetadata;
```
alisey: I think 'theme' is something more general than a style, it provides styles to multiple components. It can also provide colors, typography, icons, etc.

packages/meta-data-tools/src/types.ts
```diff
@@ -1,16 +1,35 @@
...
-export interface Simulation {
-  props?: Object;
+export interface IMetadata {
+  components: Map<ComponentType<any>, IComponentMetadata<any>>;
+  assets: Map<any, IAssetMetadata<any>>;
+  themes: Map<any, IStyleMetadata>;
 }
+
+export interface IAssetMetadata<AssetType> {
+  type: AssetType;
```
alisey: At this point AssetType is any, right? Should it be a string? We cannot store the type itself, since types don't exist at runtime.

PR #16, opened 2018-06-18
Schema - add generic support

danielstro:


packages/schema-extrct/tests/generics/classes.spec.ts
```diff
@@ -30,46 +30,47 @@ describe('schema-extrct - generic classes',()=>{
...
-                        "name":"T"
-                    },{
                         "name":"P"
+                    },{
+                        "name":"T"
                     }],
                     "arguments":[
                         {
-                            "$ref":"#typeof MyClass!T",
+                            "$ref":"#MyClass!T",
```
danielstro: This change could be a mistake, this ref seems a bit confusing so I changed it but I am not sure if the constructer arguments should be "#MyClass!T" or "#typeof MyClass!T".

packages/schema-extrct/src/json-schema-types.ts
```diff
@@ -72,36 +74,21 @@ export type ModuleSchema<T extends  SchemaTypes = SchemaTypes> = Schema<T> & {
...
-    returns:Schema;
-    properties:{[name:string]:Schema};
-    extends?:Schema
+    returns?:Schema
 }
 
 export type ClassSchema = Schema & {
     $ref:typeof ClassSchemaId;
-    constructor:Schema;
+    constructor?:FunctionSchema;
```
danielstro: We should probably decide if constructor should be under FunctionSchema or something else.

PR #25, opened 2018-06-25
Added tests and implementation for various registry features

Zombiefruit:


packages/metadata-tools/src/registry/asset-metadata.ts
```diff
@@ -0,0 +1,13 @@
+import { IAssetMetadata } from './types';
+
+export default class AssetMetadata implements IAssetMetadata {
```
alisey: How about:
```
export default class AssetMetadata implements IAssetMetadata {
  constructor(public type: string, public name: string, public description?: string) { };
}
```

packages/metadata-tools/src/registry/registry.ts
```diff
@@ -17,18 +19,24 @@ const Registry: IRegistry<IAssetMap> = {
     return this.metadata.components.get(comp)!;
   },
   describeAsset <AssetType extends keyof IAssetMap, Asset extends IAssetMap[AssetType]>(asset: Asset, type: AssetType, name: string, description?: string): IAssetMetadata {
-    // TODO: actually implement
+    if (!this.metadata.assets.has(asset)) {
```
alisey: Instead of ignoring subsequent attempts to describe an asset I think it should throw an error.

packages/metadata-tools/src/registry/index.ts
```diff
@@ -1,4 +1,5 @@
 import Registry from './registry';
 import ComponentMetadata from './component-metadata';
-export {ComponentMetadata};
+import AssetMetadata from './asset-metadata';
+export {ComponentMetadata, AssetMetadata};
```
JoMarton: check if you can use `export {default as Ninja} from './ninja` and

packages/metadata-tools/src/registry/registry.ts
```diff
@@ -1,6 +1,8 @@
 import {IRegistry, IAssetMetadata, IThemeMetadata} from './types';
 import {ComponentType} from 'react';
+import AssetMetadata from './asset-metadata';
```
JoMarton: I think its preferable to avoid default exports

packages/metadata-tools/src/registry/registry.ts
```diff
@@ -9,26 +11,32 @@ interface IAssetMap {
...
-  describeTheme(theme: any): IThemeMetadata {
-    // TODO: actually implement
+  getThemeMetadata(theme: any, name: string): IThemeMetadata {
+    if (!this.metadata.themes.has(theme)) {
+      this.metadata.themes.set(theme, new ThemeMetadata(name));
+    }
+
     return this.metadata.themes.get(theme)!;
   },
   clean() {
```
JoMarton: lets decide between `clean` and `clear` and stick to one

packages/metadata-tools/src/registry/theme-metadata.ts
```diff
@@ -0,0 +1,5 @@
+import { IThemeMetadata } from './types';
+
+export default class ThemeMetadata implements IThemeMetadata {
```
JoMarton: prefer no default exports

PR #26, opened 2018-06-26
Added behaviour to allow testing tests

Zombiefruit:


packages/auto-ssr/test/tsconfig.json
```diff
@@ -6,9 +6,12 @@
       "dom"
     ],  
     "types": [
+      "node",
```
JoMarton: node twice?

PR #28, opened 2018-06-27
Rename Portal to Modal and access DOM only in componentDidMount

alisey:


alisey:
@Zombiefruit removed the comment

PR #30, opened 2018-06-27
Added assertions to check that console.log is not called by components

Zombiefruit:


packages/sanity/src/ssr-test/ssr-test.tsx
```diff
@@ -21,10 +34,20 @@ export const ssrTest = (): void => {
                     expect(() => renderToString(<Comp />), 'RenderToString threw an error').not.to.throw();
                 });
 
+                it('should not print anything to the console', () => {
+                    renderToString(<Comp />);
+                    expect(consoleSpy.notCalled).to.equal(true);
```
alisey: Since you're using `sinonChai` you can use `expect(console.spy).not.called()` - I don't remember the exact syntax. It's fine to keep it as is, but then you don't need sinonChai.

PR #35, opened 2018-07-01
Add helpers for starting a server and configuring webpack

alisey:


packages/eyes/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    .addEntry('meta', path.join(packagePath, 'src/browser/run'))
+    .addHtml({
+      template: path.join(packagePath, 'src/browser/index.html'),
+      title: 'Eyes'
+    })
+    .suppressReactDevtoolsSuggestion()
+    .getConfig();
+}
+
+async function main() {
```
alisey: At this point this is just an example of usage, doesn't really do anything.

PR #36, opened 2018-07-01
Added linter to ensure that the only dev-dependecies are in the root

Zombiefruit:


dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
+const fs = require('fs');
+const glob = require('glob');
+
+let rootPackageJson;
+
+try {
+  rootPackageJson = JSON.parse(fs.readFileSync('./package.json'));
```
AviVahl: `'./package.json'`?

use `__dirname` and `path.join`.

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+} catch (e) {
+  throw new Error(`Failed to parse package.json, with error: ${e}`);
+}
+
+let workspaces;
```
AviVahl: why separate let?

```
const {workspaces} = rootPackageJson
if (!workspaces) { //throw }
```

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+
+if (rootPackageJson.workspaces) {
+  workspaces = rootPackageJson.workspaces;
+} else {
+  throw new Error('It seems like the root package.json doesn\'t define a workspace.');
+}
+
+
```
AviVahl: extra empty line. if this was .ts file, `tslint` would have warned.

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+} catch (e) {
+  throw new Error(`Failed to parse package.json, with error: ${e}`);
+}
+
+let workspaces;
+
+if (rootPackageJson.workspaces) {
+  workspaces = rootPackageJson.workspaces;
+} else {
+  throw new Error('It seems like the root package.json doesn\'t define a workspace.');
```
AviVahl: use template string instead of escaping the `'`.

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+
+workspaces.forEach(workspace => {
```
AviVahl: `for const of`, for readability and debuggability

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+  const packages = glob.sync(workspace);
```
AviVahl: is glob the really correct way to handle any possible string inside this key? did you check what yarn is doing?

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+  if (!packages || packages.length === 0) {
+    throw new Error('Couldn\'t find any packages in the workspace.');
```
AviVahl: template string ditto

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+  const errors = [];
+
+  packages.forEach(package => {
```
AviVahl: for const of

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+    const packageJson = JSON.parse(fs.readFileSync(`${package}/package.json`));
```
AviVahl: - try catch
- path.join instead of string concat

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+    if (packageJson.devDependencies) {
+      for (const devDep in packageJson.devDependencies) {
+        errors.push(`\n Package "${package}" cannot have devDependency "${devDep}" in its package.json. This dev-dependency should be placed in the root package.json`);
```
AviVahl: if errors is array, why push `\n`? eventually you should `.join('\n')`
"dev-dependency"...

dev-dependency-linter.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+  })
+
+  if (errors.length > 0) {
+    throw new Error(errors.toString());
```
AviVahl: errors.join('\n')

package.json
```diff
@@ -9,7 +9,7 @@
     "build": "lerna run build --stream",
     "build:watch": "lerna run --parallel build:watch",
     "pretest": "run-p lint build",
-    "lint": "lerna run lint --stream",
+    "lint": "lerna run lint --stream && node dev-dependency-linter",
```
AviVahl: use `./<whatever path>` when running directly with node.

PR #40, opened 2018-07-04
Accessibility tests

danielstro:


packages/a11y/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "a11y",
```
alisey: If we plan to publish these packages I think we need to have the same prefix for all of them.`ui-autotools-a11y`?

packages/utils/src/puppeteer.ts
```diff
@@ -0,0 +1,13 @@
+import puppeteer from 'puppeteer';
+
+export function waitForPageError(page: puppeteer.Page): Promise<never> {
+    return new Promise((_, reject) => {
+        page.on('pageerror', (errorText: string) => {
+        reject(new Error(errorText));
```
alisey: Lost indentation here.

packages/autotools-scripts/src/cli.ts
```diff
@@ -17,6 +18,18 @@ program
   ssrTest();
 });
 
+program
+.command('a11y')
+.description('test')
+.option('-p, --path <p>', 'project path')
+.option('-i, --impact <i>', 'Only display issues with impact level higher than <i>. Value between 1 (minor) and 4 (critical)')
```
alisey: I think it would be easier for the users to specify this option as a string: `--impact minor`. Then we also wouldn't need a to convert it from number to string and back.

packages/autotools-scripts/src/cli.ts
```diff
@@ -17,6 +18,18 @@ program
   ssrTest();
 });
 
+program
+.command('a11y')
+.description('test')
+.option('-p, --path <p>', 'project path')
```
alisey: Do we need the project path? We can use `process.cwd()` that gives the working directory (which is the project directory). And all tools should probably support `--files` option which allows to control which subset of metafiles is loaded.

packages/a11y/src/browser/run.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const root = document.getElementById('react-root');
+async function test(rootElement: HTMLElement) {
+  const comps = createTestsFromSimulations(root);
+  for (const c of comps) {
+    const div = document.createElement('div');
+    rootElement.appendChild(div);
+    await c.render(div);
+  }
+  axe.run(rootElement, (err: Error, result: axe.AxeResults) => {
```
alisey: It would we very helpful if the errors contained information about which component and which simulation has lead to the error. One way to do that is to run tests one by one (unless this turns out to be really slow).

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      title: 'Accessibility'
+    })
+    .suppressReactDevtoolsSuggestion()
+    .getConfig();
+}
+
+function printResults(results: axe.AxeResults, impact: number): string {
+  const msg: string[] = [];
+  let index = 1;
+  results.violations.forEach((violation) => {
```
alisey: A second argument to forEach gives you the index, so it's not really necessary to define the index variable and increment it manually.

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (impactArray.indexOf(violation.impact) + 1 >= impact) {
+      violation.nodes.forEach((node) => {
+        msg.push(`${index++}. ${chalk.red(violation.id === 'duplicate-id' ? 'Document' : node.target[0])}: (Impact: ${violation.impact}) ${node.failureSummary}`);
```
alisey: This would be easier to read if it used intermediate variables to build up the message, which would also help with understanding what is `duplicate-id` and `node.target[0]`

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    .addEntry('meta', path.join(ownPath, 'esm/browser/run'))
+    .addHtml({
+      template: path.join(ownPath, 'src/browser/index.html'),
+      title: 'Accessibility'
+    })
+    .suppressReactDevtoolsSuggestion()
+    .getConfig();
+}
+
+function printResults(results: axe.AxeResults, impact: number): string {
```
alisey: Doesn't really print the results, just formats them. `formatResults`?

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  let index = 1;
+  results.violations.forEach((violation) => {
+    if (impactArray.indexOf(violation.impact) + 1 >= impact) {
+      violation.nodes.forEach((node) => {
+        msg.push(`${index++}. ${chalk.red(violation.id === 'duplicate-id' ? 'Document' : node.target[0])}: (Impact: ${violation.impact}) ${node.failureSummary}`);
+      });
+    }
+  });
+  if (msg.length === 0) {
+    return 'No errors found';
```
alisey: In the case of success we don't really need to print anything unless we're in some kind of verbose mode.

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  } else {
+    return msg.join('\n\n');
+  }
+}
+
+export async function a11yTest(p: string, impact: number) {
```
alisey: Not clear what `p` is.

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+    await page.goto(server.getUrl());
+    const results = await Promise.race([waitForPageError(page), getResults]);
+    console.log(printResults(results, impact));
+
+    process.exitCode = 0;
+  } catch (error) {
+    if (browser) {
+      try {
+       browser!.close();
```
alisey: Better to close both server and browser in `finally`

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    const page = await browser.newPage();
+    const getResults = new Promise<axe.AxeResults>((resolve) =>
+      page.exposeFunction('runAxeTest', resolve)
+    );
+    page.on('dialog', (dialog) => {
+      dialog.dismiss();
+    });
+    await page.goto(server.getUrl());
+    const results = await Promise.race([waitForPageError(page), getResults]);
+    console.log(printResults(results, impact));
```
alisey: The process needs to quit with `exitCode=1` in the case of violation. Otherwise CI will think that the tests succeeded.

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,75 @@
+/* tslint:disable:no-console */
```
Zombiefruit: It would be nice to implement a `log` method in utils so that we only have to disable console in one place. @AviVahl showed me this the other day

packages/a11y/src/browser/run.tsx
```diff
@@ -23,18 +31,21 @@ function createTestsFromSimulations(reactRoot: any) {
 
 const root = document.getElementById('react-root');
 async function test(rootElement: HTMLElement) {
+  const results: IResult[] = [];
   const comps = createTestsFromSimulations(root);
```
alisey: `tests` is a more suitable name.

packages/a11y/src/browser/run.tsx
```diff
@@ -23,18 +31,21 @@ function createTestsFromSimulations(reactRoot: any) {
...
   for (const c of comps) {
-    const div = document.createElement('div');
-    rootElement.appendChild(div);
-    await c.render(div);
-  }
-  axe.run(rootElement, (err: Error, result: axe.AxeResults) => {
-    if (err) {
-      throw err;
+    try {
+      const div = document.createElement('div');
```
alisey: Is this necessary, or maybe we could render directly into the root?

packages/a11y/src/browser/run.tsx
```diff
@@ -23,18 +31,21 @@ function createTestsFromSimulations(reactRoot: any) {
...
+      const result = await axe.run(rootElement);
+      results.push({comp: c.title, result});
+      await c.cleanup();
+    } catch (error) {
+      results.push({comp: c.title, error});
     }
-    (window as any).runAxeTest(result);
-  });
+  }
+  (window as any).runAxeTest(results);
```
alisey: `puppeteerReportResults` or `sendResultsToPuppeteer`, or something like that.

packages/a11y/src/index.ts
```diff
@@ -21,52 +23,58 @@ function getWebpackConfig(p: string) {
...
-      violation.nodes.forEach((node) => {
-        msg.push(`${index++}. ${chalk.red(violation.id === 'duplicate-id' ? 'Document' : node.target[0])}: (Impact: ${violation.impact}) ${node.failureSummary}`);
+  results.forEach((res) => {
+    if (res.error) {
+      msg.push(`${index++}. ${res.comp}: Error while testing component - ${res.error}`);
+    } else if (res.result) {
+      res.result.violations.forEach((violation) => {
+        if (impactArray.indexOf(violation.impact) + 1 >= impact) {
+          violation.nodes.forEach((node) => {
+            const compName = (`${res.comp} - ${node.target[0]}`);
```
alisey: It's not clear what `node.target[0]` is, and giving it a name would help. Something like `const cowsPerCubicMeter = node.target[0]`.

Is it possible that there's going to be 0 or more than one targets?

packages/a11y/src/index.ts
```diff
@@ -21,52 +23,58 @@ function getWebpackConfig(p: string) {
...
-  }
+  return msg.join('\n\n');
 }
 
-export async function a11yTest(p: string, impact: number) {
+export async function a11yTest(entry: string | string[], impact: number) {
   let server: IServer | null = null;
   let browser: puppeteer.Browser | null = null;
-  console.log('Running accessibility tests...');
+  consoleLog('Running accessibility tests...');
```
alisey: I think that by default our commands should print only errors. npm prints the name of the script it's running anyway, so this is redundant.

packages/a11y/src/index.ts
```diff
@@ -21,52 +23,58 @@ function getWebpackConfig(p: string) {
...
     });
     await page.goto(server.getUrl());
     const results = await Promise.race([waitForPageError(page), getResults]);
-    console.log(printResults(results, impact));
-
-    process.exitCode = 0;
+    const message = formatResults(results, impact);
+    if (message) {
+      process.exitCode = 1;
+      consoleLog(message);
```
alisey: `process.stderr.write(message)`

packages/autotools-scripts/src/cli.ts
```diff
@@ -21,13 +26,13 @@ program
 program
 .command('a11y')
 .description('test')
-.option('-p, --path <p>', 'project path')
+.option('-f, --files [pattern]', 'Grep file')
 .option('-i, --impact <i>', 'Only display issues with impact level higher than <i>. Value between 1 (minor) and 4 (critical)')
 .action((options) => {
+  const entry = glob.sync(path.join(packagePath, options.files ? options.files : defaultMetaGlob));
```
alisey: This is going to look for meta files in the `a11y` directly, which is certainly wrong.

packages/utils/src/webpack.ts
```diff
@@ -12,6 +12,16 @@ export class WebpackConfigurator {
     return this.config;
   }
 
+  public setEntry(name: string, entry: string | string[]): this {
```
alisey: This function should set entry to a specific value, it shouldn't append to existing one.

packages/a11y/src/browser/run.tsx
```diff
@@ -9,7 +9,7 @@ interface ITest {
   cleanup: () => void;
 }
 
-export interface IResult {
+export interface ITests {
```
alisey: I think `IResult` was more suitable.

packages/a11y/src/browser/run.tsx
```diff
@@ -18,34 +18,32 @@ export interface IResult {
...
+    }
   }
   return tests;
 }
 
 const root = document.getElementById('react-root');
 async function test(rootElement: HTMLElement) {
-  const results: IResult[] = [];
+  const results: ITests[] = [];
   const comps = createTestsFromSimulations(root);
```
alisey: There are not components, these are tests.

packages/a11y/src/index.ts
```diff
@@ -46,13 +46,12 @@ function formatResults(results: IResult[], impact: number): string {
 export async function a11yTest(entry: string | string[], impact: number) {
   let server: IServer | null = null;
   let browser: puppeteer.Browser | null = null;
-  consoleLog('Running accessibility tests...');
   try {
     server = await serve({webpackConfig: getWebpackConfig(entry)});
     browser = await puppeteer.launch({headless: true});
```
alisey: It's headless by default, don't need to pass the option.

packages/a11y/src/browser/run.tsx
```diff
@@ -0,0 +1,49 @@
...
+  cleanup: () => void;
+}
+
+export interface IResult {
+  comp: string;
+  result?: axe.AxeResults;
+  error?: Error;
+}
+
+function createTestsFromSimulations(reactRoot: any) {
```
alisey: any -> HTMLElement

packages/a11y/src/browser/run.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const tests: ITest[] = [];
+  for (const [Comp, meta] of Registry.metadata.components.entries()) {
+    for (const [simIndex, sim] of meta.simulations.entries()) {
+      tests.push({
+        title: (Comp.displayName ? Comp.displayName : Comp.name) + ' ' + simIndex,
+        render:  (container: HTMLElement) => ReactDOM.render(<div id={Comp.name}><Comp {...sim.props} /></div>, container),
```
alisey: Probably don't need the div? `<div id={Comp.name}>`

packages/a11y/src/index.ts
```diff
@@ -0,0 +1,83 @@
+import path from 'path';
+import puppeteer from 'puppeteer';
+import {WebpackConfigurator, serve, IServer, waitForPageError, consoleError} from 'ui-autotools-utils';
+import { IResult } from './browser/run';
+import axe from 'axe-core';
+import chalk from 'chalk';
+
+const ownPath = path.resolve(__dirname, '..');
+export const impactArray: axe.ImpactValue[] = ['minor', 'moderate', 'serious', 'critical'];
```
alisey: `impactArray` -> `impactLevels`

packages/autotools-scripts/src/cli.ts
```diff
@@ -3,20 +3,37 @@ import {Command} from 'commander';
...
 });
 
+program
+.command('a11y')
+.description('run accessibility tests on components with metadata files that match the given pattern')
+.option('-f, --files [pattern]', 'metadata file pattern')
+.option('-i, --impact <i>', `Only display issues with impact level <i> and higher. Values are: ${impactArray.join(', ')}`)
+.action((options) => {
+  const entry = glob.sync(path.join(projectPath, options.files ? options.files : defaultMetaGlob));
+  const impact = impactArray.indexOf(options.impact) > -1 ? options.impact : 'minor';
```
alisey: Should throw or log an error if the level is not supported.

PR #42, opened 2018-07-15
Added hydration tests to Sanity

Zombiefruit:


packages/sanity/src/ssr-renderer/run-in-puppeteer.ts
```diff
@@ -0,0 +1,94 @@
...
+    browser = await puppeteer.launch({headless: true, args});
+    const page = await browser.newPage();
+    await page.setViewport({width: viewportWidth, height: viewportHeight});
+
+    page.on('dialog', (dialog) => {
+      dialog.dismiss();
+    });
+
+    page.on('console', async (msg) => {
+      // tslint:disable-next-line:no-shadowed-variable
```
danielstro: Is this needed?

packages/sanity/src/ssr-renderer/run-in-puppeteer.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export async function runTestsInPuppeteer({testPageUrl, noSandbox}: any) {
+  const loadTimeout = 20000;
+  const testTimeout = 5000;
+  const viewportWidth = 800;
+  const viewportHeight = 600;
+
+  let browser;
+  try {
```
danielstro: Add catch

packages/sanity/src/ssr-renderer/test-page.ts
```diff
@@ -0,0 +1,30 @@
+import chai from 'chai';
+// import {patchConsole} from 'ui-autotools-utils';
```
danielstro: Remove

package.json
```diff
@@ -64,9 +66,7 @@
     "typescript": "~2.9.2",
     "typescript-support": "^0.4.1",
     "url-loader": "^1.0.1",
-    "webpack": "^4.12.1",
-    "webpack-cli": "^3.0.8",
-    "webpack-dev-server": "^3.1.4"
+    "webpack": "^4.12.1"
```
alisey: should be a normal dependency in sanity or utils or both

packages/sanity/src/ssr-renderer/test-page.ts
```diff
@@ -0,0 +1,28 @@
+import chai from 'chai';
+(window as any).expect = chai;
+
+// Mocha officially supports these two imports in browser environment.
```
alisey: There's only one import here.

packages/sanity/src/ssr-test/hydrate-tests.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+chai.use(sinonChai);
+
+export const hydrateTests = (): void => {
+  describe('Hydration tests', () => {
+    let consoleSpy: sinon.SinonSpy;
+    let errorSpy: sinon.SinonSpy;
+    let index = 0;
+    const componentStrings = (window as any).components;
+    const root = document.getElementById('root');
```
alisey: Cast it to HTMLElement here.

packages/sanity/src/ssr-test/hydrate-tests.tsx
```diff
@@ -0,0 +1,51 @@
...
+          afterEach(() => {
+              consoleSpy.restore();
+              errorSpy.restore();
+          });
+
+          it(`should hydrate ${Comp.name} with props ${JSON.stringify(simulation)} without errors`, () => {
+            const consoleArgs = consoleSpy.getCall(0) ? consoleSpy.getCall(0).args[0] : '';
+            const errorArgs = errorSpy.getCall(0) ? errorSpy.getCall(0).args[0] : '';
+              // tslint:disable-next-line:no-unused-expression
+            expect(consoleSpy, `console was called with:\n ${consoleArgs}`).to.not.be.called;
```
alisey: It it possible to use `.to.not.be.called()`? If that works you don't need tslint:disable.

packages/sanity/src/ssr-test/hydrate-tests.tsx
```diff
@@ -0,0 +1,51 @@
...
+              index++;
+          });
+
+          afterEach(() => {
+              consoleSpy.restore();
+              errorSpy.restore();
+          });
+
+          it(`should hydrate ${Comp.name} with props ${JSON.stringify(simulation)} without errors`, () => {
+            const consoleArgs = consoleSpy.getCall(0) ? consoleSpy.getCall(0).args[0] : '';
```
alisey: Does it give you just the first argument or all of them?

packages/sanity/src/ssr-test/hydrate-tests.tsx
alisey: `${consoleArgs}` - wouldn't it just give you `[[object Object]]` if one of the args is not a primitive value?

packages/sanity/src/ssr-test/hydrate-tests.tsx
```diff
@@ -0,0 +1,51 @@
...
+    let consoleSpy: sinon.SinonSpy;
+    let errorSpy: sinon.SinonSpy;
+    let index = 0;
+    const componentStrings = (window as any).components;
+    const root = document.getElementById('root');
+
+    Registry.metadata.components.forEach((metadata, Comp) => {
+      describe(Comp.name, () => {
+        metadata.simulations.forEach(((simulation) => {
+          beforeEach(() => {
```
alisey: Don't really see a reason why this entire block shouldn't be in the test. The control flow would be more linear.

packages/sanity/src/ssr-renderer/test-page.ts
```diff
@@ -0,0 +1,28 @@
+import chai from 'chai';
+(window as any).expect = chai;
```
alisey: Do we need this? In the tests you're importing it explicitly.

packages/sanity/src/ssr-test/strict-mode-tests.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
```
alisey: Do you plan to combine strict mode tests and hydration tests? The files look almost identical.

packages/sanity/src/ssr-test/strict-mode.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
```
alisey: Looks like a copy of the previous file.

packages/utils/patch-console.js
```diff
@@ -0,0 +1,334 @@
+// Based on an old version of util.inspect from Node.
+// In headless mode we run Mocha with console reporter and pipe all console
+// output into stdout. But passing structures - especially circular - between
+// browser and Node is inefficient and can cause Puppeteer to freeze. The
+// solution is to wrap console.log with a function that does all formatting
+// in the browser. This allows us to pass only primitive values to Node.
+// __HEADLESS__ is a boolean value injected by Webpack.
```
alisey: This file doesn't use __HEADLESS__

packages/utils/patch-console.js
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  return braces[0] + base + ' ' + output.join(', ') + ' ' + braces[1];
+}
+
+function isPrimitive(x) {
+  return Object(x) !== x;
+}
+
+(function() {
```
alisey: You could wrap this entire file in a function and export it. This would 1) prevent the functions from being exposed globally, and 2) allow you to import the file normally instead of using readFileSync.
```
import {patchConsole} from 'blabla';
page.evaluateOnNewDocument(patchConsole);
```

packages/utils/src/puppeteer.ts
```diff
@@ -11,3 +16,106 @@ export function waitForPageError(page: puppeteer.Page): Promise<never> {
...
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function loadTestPage(page: any, testPageUrl: string, timeout: number) {
+  // This can keep the process from terminating for upto `timeout` if an error
+  // occurs on the page before page load event. But the problem should not occur
+  // in headless mode.
+  // Bug: https://github.com/GoogleChrome/puppeteer/issues/2721
+  try {
```
alisey: What's the reason for try catch?

packages/utils/src/puppeteer.ts
```diff
@@ -11,3 +16,106 @@ export function waitForPageError(page: puppeteer.Page): Promise<never> {
         });
     });
 }
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function loadTestPage(page: any, testPageUrl: string, timeout: number) {
```
alisey: `page: any` -> `puppeteer.Page`

packages/utils/src/puppeteer.ts
```diff
@@ -11,3 +16,106 @@ export function waitForPageError(page: puppeteer.Page): Promise<never> {
...
+    const args = noSandbox ? ['--no-sandbox', '--disable-setuid-sandbox'] : [];
+    browser = await puppeteer.launch({headless: true, args});
+    const page = await browser.newPage();
+    await page.setViewport({width: viewportWidth, height: viewportHeight});
+
+    page.on('dialog', (dialog) => {
+      dialog.dismiss();
+    });
+
+    page.on('console', async (msg) => {
```
alisey: Thought we already have this somewhere in the helpers.

packages/utils/patch-console.js
```diff
@@ -0,0 +1,334 @@
...
+    }
+  
+    return braces[0] + base + ' ' + output.join(', ') + ' ' + braces[1];
+  }
+  
+  function isPrimitive(x) {
+    return Object(x) !== x;
+  }
+  
+  (function() {
```
alisey: Don't need this IIFE now.

packages/utils/patch-console.js
```diff
@@ -0,0 +1,334 @@
+// Based on an old version of util.inspect from Node.
+// In headless mode we run Mocha with console reporter and pipe all console
+// output into stdout. But passing structures - especially circular - between
+// browser and Node is inefficient and can cause Puppeteer to freeze. The
+// solution is to wrap console.log with a function that does all formatting
+// in the browser. This allows us to pass only primitive values to Node.
+module.exports.patchConsole = function() {
```
alisey: `exports.patchConsole = ...` would work.

packages/utils/src/puppeteer.ts
```diff
@@ -11,3 +13,102 @@ export function waitForPageError(page: puppeteer.Page): Promise<never> {
...
+        consoleLog(...msgArgs);
+    });
+}
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function loadTestPage(page: puppeteer.Page, testPageUrl: string, timeout: number) {
+  // This can keep the process from terminating for upto `timeout` if an error
```
alisey: Remove this part: `But the problem should not occur in headless mode.`, I later discovered that it does occur in headless mode, just less frequently.

packages/sanity/src/ssr-test/sanity-tests.tsx
```diff
@@ -0,0 +1,68 @@
...
+            errorSpy = sinon.spy(console, 'error');
+          });
+
+          afterEach(() => {
+            consoleSpy.restore();
+            errorSpy.restore();
+          });
+
+          it(`should render ${Comp.name} in strict mode with props ${JSON.stringify(simulation)} without errors`, () => {
+            // Hydrate the component
```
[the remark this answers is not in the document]
Zombiefruit: Indeed it isn't

packages/utils/src/puppeteer.ts
```diff
@@ -11,3 +13,101 @@ export function waitForPageError(page: puppeteer.Page): Promise<never> {
...
+      loadTestPage(page, testPageUrl, loadTimeout).then(() =>
+        Promise.race([
+          waitForTestResults(page),
+          failIfTestsStall(page, testTimeout)
+        ])
+      )
+    ]);
+
+    return numFailedTests;
+  } catch (error) {
```
alisey: Just one last thing, this function should throw an exception in the case of error instead of simply logging the result. I think you can simply remove the `catch` block and everything should work fine.

PR #44, opened 2018-07-18
Schema Linker

danielstro:
Added linker and fixed a small issue in schema-extract with parentheses types.


packages/schema-extract/src/file-linker.ts
```diff
@@ -0,0 +1,192 @@
...
+        if (isRef(option)) {
+            let entity: Schema & IObjectFields;
+            if (paramsMap) {
+                entity = paramsMap!.get(option.$ref)!;
+                if (!entity) {
+                    return res;
+                }
+            } else {
+                const refEntity = option.genericArguments ? option : schema.definitions![option.$ref!.replace('#', '')];
+                entity = link(refEntity, schema);
```
nadavwix: if only one of them has it, the result does not
if both have it, , the result should have an index signature thats the intersection of both.

its worth to note that index signature for known keys i.e.

type A = 'a' | 'b'
type B = 'a' | 'c'
type A1 = {
    [key:string]:string
}
type B1 = {
    [key in B]:"gaga"
}
type c = A1 & B1;

should be linked so that c is treated as

type c = {
    a:"gaga",
    b:string,
    c:"gaga"
}

packages/schema-extract/test/linker/intersections.spec.ts
```diff
@@ -0,0 +1,196 @@
+import {expect} from 'chai';
+import { Schema } from '../../src/json-schema-types';
+import {linkTest} from '../../test-kit/run-linker';
+
+describe('schema-linker - intersections', () => {
+    it('should flatten intersection types', async () => {
+        const fileName = 'index.ts';
```
nadavwix: add test for: 
```
type a {
   something:string
}

type b {
   something:number
}
```

and

```
type a {
   something:string
}

type b {
   something: gaga
}
```

PR #46, opened 2018-07-19
Fixing sanity/hydration tests

Zombiefruit:


packages/sanity/src/hydration-test/server.tsx
```diff
@@ -1,33 +1,31 @@
...
 
-function getWebpackConfig(ssrComps: string[]) {
+function getWebpackConfig(ssrComps: string[], metaGlob: string) {
   return WebpackConfigurator
     .load(webpackConfigPath)
-    .setEntry('meta', glob.sync(path.join(projectDir, metaGlob)))
-    .addEntry('meta', path.join(packageDir, 'ssr-renderer', 'test-page.js'))
-    .addEntry('meta', path.join(packageDir, 'ssr-test', 'index.js'))
+    .setEntry('meta', glob.sync(metaGlob))
+    .addEntry('meta', path.join(packageDir, 'hydration-test', 'test-page.js'))
```
alisey: Couldn't you require the test page in the index file instead of adding it as an entry?

PR #60, opened 2018-07-24
Made name or displayName property mandatory

Zombiefruit:


alisey:
Created a couple of cards for this.

PR #63, opened 2018-07-24
Made import more specific to avoid importing browser incompatible code

Zombiefruit:


alisey:
@AviVahl do you have suggestions about how to avoid this pattern of explicitly specifying cjs/esm in our imports?

PR #67, opened 2018-07-26
Verifying that simulation and component names are alphanumeric

Zombiefruit:


packages/registry/src/utils/verify-string.ts
```diff
@@ -0,0 +1,5 @@
+export function isAlphanumeric(stringToVerify: string): boolean {
```
alisey: Alphanumeric means consisting of letters and numbers, but I have no suggestions for a better name. Maybe we should have `isValidSimulationTitle` and `isValidComponentName`, and the latter will have a different validation rule that only allows letters and numbers, and no spaces or punctuation. What do you think?

packages/registry/test/fixtures/component-fixtures.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
```
alisey: import React from 'react';

packages/registry/test/utils-tests/verify-string.spec.ts
```diff
@@ -0,0 +1,32 @@
+import {isValidSimulationTitle, isValidComponentName} from '../../src';
+import {expect} from 'chai';
+
+describe('Verification Methods', () => {
+  describe('isValidSimulationTitle', () => {
+    it('returns true for an valid string', () => {
+      const goodString = 'TestMe123';
```
danielstro: Maybe we should use all acceptable character types in the valid string?

PR #68, opened 2018-07-29
Added screenshot utility

Zombiefruit:


packages/eyes/src/index.ts
```diff
@@ -1,16 +1,39 @@
...
+import {WebpackConfigurator, serve, IServer, waitForPageError, logConsoleMessages, consoleLog} from '@ui-autotools/utils';
+import chalk from 'chalk';
+import uuid from 'uuid';
 
 const packagePath = path.resolve(__dirname, '..');
 const projectPath = process.cwd();
 const webpackConfigPath = path.join(projectPath, 'meta.webpack.config.js');
 
-function getWebpackConfig() {
+const projectName = require(path.join(projectPath, 'package.json')).name;
```
alisey: Let's also validate this in case someone accidentally runs the script in a wrong folder.

packages/eyes/src/index.ts
```diff
@@ -1,16 +1,39 @@
...
+const batchName   = projectName;
+const batchId     = uuid.v4();
+const eyesApiKey  = process.env.EYES_API_KEY;
+const osName      = process.platform;
+
+if (!eyesApiKey) {
+  process.exitCode = 1;
+  throw new Error('"EYES_API_KEY" needs to be defined in a ".env" file located in the root.');
```
alisey: just say that this environment variable needs to be defined, in CI it gets defined through the 
 admin console, for example. `.env` is just one of the options.

packages/eyes/src/index.ts
```diff
@@ -58,6 +81,20 @@ async function runTests(url: string) {
   }
 }
 
+function logEyesResult({name, status, isNew, appUrls}: any) {
+  const isError = status === 'Unresolved' && isNew.toString() === 'false';
```
alisey: Isn't isNew a boolean? I think isError should be status !== 'Passed' instead of status === 'Unresolved'.

packages/eyes/src/index.ts
```diff
@@ -72,21 +109,34 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
 
   // TODO: add a timeout for each test to make sure we terminate if they get
   // stuck.
+  let testsPassed: boolean = true;
+  consoleLog('Format: <component simName>');
```
alisey: What's this?

packages/eyes/src/index.ts
```diff
@@ -72,21 +109,34 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
...
   for (const [i, {title}] of tests.entries()) {
+    let result;
     await page.evaluate(`puppeteerRenderTest(${i})`);
     const screenshot = await page.screenshot();
-    // TODO: Compare the screenshot using eyes
-    console.log({title, screenshotBufferSize: screenshot.byteLength});
+    await eyes.open(projectName, title, null);
+    const {asExpected} = await eyes.checkImage(screenshot);
```
alisey: What does checkImage return in the case of a new image?

packages/eyes/src/index.ts
```diff
@@ -72,21 +109,34 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
...
+    if (!asExpected && testsPassed) {
+      testsPassed = false;
+    }
+    result = await eyes.close(false);
+    logEyesResult(result);
     await page.evaluate(`puppeteerCleanupTest(${i})`);
   }
 
+  if (!testsPassed) {
```
alisey: Setting status/quitting should be in the top level function. This function should simply return the number of failed tests.

packages/eyes/src/index.ts
```diff
@@ -58,6 +81,20 @@ async function runTests(url: string) {
...
+function logEyesResult({name, status, isNew, appUrls}: any) {
+  const isError = status === 'Unresolved' && isNew.toString() === 'false';
+  const dividerString = '--------------------';
+  const errorString = '~~~~~~~~~~~~~~~~~~~~';
+  const divider = isError ? chalk.red(errorString) : dividerString;
+
+  consoleLog(`
+  ${divider}
+  ${isError ? chalk.red('ERROR') : ''} ${chalk.bold(name)}: ${status === 'Unresolved' ? chalk.red(status) : chalk.green(status)}. Is new: ${isNew.toString() === 'false' ? chalk.green(isNew) : chalk.yellow(isNew)}.
+  ${chalk.cyan('URL')}: ${chalk.underline(appUrls.batch)}
```
alisey: We don't need to print the batch URL for each failed test, let's print it once at the bottom.

packages/eyes/src/index.ts
```diff
@@ -58,6 +81,20 @@ async function runTests(url: string) {
   }
 }
 
+function logEyesResult({name, status, isNew, appUrls}: any) {
+  const isError = status === 'Unresolved' && isNew.toString() === 'false';
+  const dividerString = '--------------------';
+  const errorString = '~~~~~~~~~~~~~~~~~~~~';
+  const divider = isError ? chalk.red(errorString) : dividerString;
+
+  consoleLog(`
```
alisey: Really hard to read, maybe split handling of new / mismatch into two branches?

packages/eyes/src/index.ts
Zombiefruit: `asExpected` is false in this case, so I added a check to ensure that the component isn't new

packages/eyes/src/index.ts
```diff
@@ -1,16 +1,43 @@
-/* tslint:disable:no-console */
+require('dotenv').config();
```
alisey: I think it should be the project's responsibility to populate environment variables. We shouldn't expect that the project uses `.env`, and should simply read `process.env`

packages/eyes/src/index.ts
```diff
@@ -58,6 +85,11 @@ async function runTests(url: string) {
   }
 }
 
+function logEyesResult(isError: boolean, {name, status, isNew, appUrls}: any) {
+  const url = isError ? `${chalk.cyan('URL')}: ${chalk.underline(appUrls.session)}` : '';
+  consoleLog(`${isError ? chalk.red('👎 FAIL') : isNew ? chalk.yellow('👌  NEW') : chalk.green('👍  OK')} ${chalk.bold(name)}. ${url}`);
```
alisey: Too many things are happening on the same line.

```
const status = isError ? chalk.red('👎 FAIL') : 
               isNew ? chalk.yellow('👌  NEW') :
               chalk.green('👍  OK');

consoleLog(`${status} ${chalk.bold(name)}. ${url}`);
```

packages/eyes/src/index.ts
```diff
@@ -58,6 +85,11 @@ async function runTests(url: string) {
   }
 }
 
+function logEyesResult(isError: boolean, {name, status, isNew, appUrls}: any) {
```
alisey: We're not using status now and can remove it.

packages/eyes/src/index.ts
```diff
@@ -72,21 +104,33 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
 
   // TODO: add a timeout for each test to make sure we terminate if they get
   // stuck.
+  let numTestsFailed: number = 0;
```
alisey: Do we have a card for the timeout task?

PR #69, opened 2018-07-29
Linker - Added import linking

danielstro:


packages/schema-extract/src/extract-schema.ts
```diff
@@ -4,18 +4,31 @@ import typescript from 'typescript';
 import {LocalFileSystem} from 'kissfs';
 import {transform} from './file-transformer';
 import {createHost} from './isomorphc-typescript-host';
+import {SchemaLinker} from './file-linker';
 
-export function* extractSchema(basePath: string, filesGlob: string) {
+export function* extractSchema(basePath: string, filesGlob: string, linked = false) {
```
nadavwix: the linked schema should be moved to a different function.

it requires the name of the export to link

packages/schema-extract/src/file-linker.ts
```diff
@@ -1,140 +1,193 @@
...
+        this.sourceFile = sourceFile;
+        const schema = transform(this.checker, sourceFile, file, this.projectPath);
+        let entity;
+        if (schema.definitions) {
+            entity = schema.definitions[entityName];
+        } else if (schema.properties) {
+            entity = schema.properties[entityName];
+        }
         if (!entity) {
             return {};
```
nadavwix: should return common/unknown

packages/schema-extract/test/linker/imports.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+                something: {
+                    type: 'string'
+                }
+            },
+            required: ['something']
+        };
+        expect(res).to.eql(expected);
+    });
+
+    it('should link imported type from an outside package', async () => {
```
nadavwix: should add a test that shows we work on a linked d.ts file

packages/schema-extract/src/extract-schema.ts
danielstro: I created a new function and changed it back. Anyway the cli command doesn't work properly yet and I will fix it separately

packages/schema-extract/src/file-linker.ts
danielstro: Fixed

PR #73, opened 2018-07-31
Moved the env variable check for the eyes api key to the cli tool 

Zombiefruit:


packages/autotools-scripts/src/cli.ts
```diff
@@ -48,6 +48,12 @@ program
 .option('-f, --files [pattern]', 'metadata file pattern')
 .action((options) => {
   const entry = glob.sync(path.join(projectPath, options.files ? options.files : defaultMetaGlob));
+
+  if (!process.env.EYES_API_KEY) {
```
alisey: I meant this should be in the `eyesTest` function. Another option is to pass `eyesApiKey` as a param to `eyesTest()`, and then it would be validated inside of that function.

Also if we're going to do `require('dotenv').config();` we should do it in this file.

packages/eyes/src/index.ts
```diff
@@ -124,8 +123,13 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
   return numTestsFailed;
 }
 
-export async function eyesTest(entry: string | string[]) {
+export async function eyesTest(entry: string | string[], apiKey: string) {
   let server: IServer | null = null;
+  if (!apiKey) {
+    process.exitCode = 1;
+    throw new Error('The environment variable "EYES_API_KEY" needs to be defined.');
```
alisey: It's not going to get to the next line if you throw here.

packages/eyes/src/index.ts
```diff
@@ -131,11 +105,33 @@ async function waitForTestsCompletion(page: puppeteer.Page, url: string):
   return numTestsFailed;
 }
 
-export async function eyesTest(entry: string | string[]) {
+export async function eyesTest(entry: string | string[], apiKey: string, projectPath: string, webpackConfigPath: string) {
   let server: IServer | null = null;
+  const projectName = require(path.join(projectPath, 'package.json')).name;
```
alisey: This line can throw an exception, and since this is an async function Node will terminate with status 0, as if the tests have passed successfully.

I suggest that you move the beginning of that try block on line 132 to the beginning of the function, and remove process.exit in apiKey and projectName validations.

PR #90, opened 2018-08-05
Website: command for building a static website

alisey:


packages/website/src/server/index.ts
```diff
@@ -13,6 +14,24 @@ import {getMetadataAndSchemasInDirectory} from './meta';
...
 const StylableWebpackPlugin = require('stylable-webpack-plugin');
 
+interface IProjectOptions {
+  projectPath: string;
+  webpackConfigPath: string;
+  metadataGlob: string;
+  sourcesGlob: string;
+}
+
+export interface IStartWebisteOptions {
```
danielstro: typo in 'webiste'

packages/website/src/server/index.ts
```diff
@@ -13,6 +14,24 @@ import {getMetadataAndSchemasInDirectory} from './meta';
...
+  projectOptions: IProjectOptions;
+  host: string;
+  port: number;
+}
+
+export interface IBuildWebisteOptions {
```
danielstro: same typo

packages/website/src/server/index.ts
```diff
@@ -28,17 +47,20 @@ function getServerUrl(server: http.Server) {
 }
 
 function getWebsiteWebpackConfig(
-  options: {
-    projectPath: string;
+  {
```
danielstro: This is a bit weird, do you need to give it a type? can you maybe just create a type alias for readability?

packages/website/src/server/index.ts
```diff
@@ -73,26 +95,30 @@ function getWebsiteWebpackConfig(
 }
 
 function getMetadataWebpackConfig(
-  options: {
-    projectPath: string;
-    metadataGlob: string;
+  {
```
danielstro: same comment as above

packages/website/src/server/index.ts
alisey: Could you clarify what you mean?

packages/website/src/server/index.ts
```diff
@@ -152,12 +168,59 @@ export async function startWebsite(options: IStartWebisteOptions) {
...
+    // in the project it's easier to let Webpack catch and report them instead
+    // of implementing friendly error reporting in the schema extractor.
+
+    const metadataCompiler = webpack(getMetadataWebpackConfig({
+      projectOptions,
+      production: true,
+      outputPath
+    }));
+
+    await new Promise((resolve, reject) => {
```
JoMarton: this could be a good case for `promisify` https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original

packages/website/src/server/index.ts
```diff
@@ -152,12 +168,59 @@ export async function startWebsite(options: IStartWebisteOptions) {
...
+      });
+    });
+
+    const websiteCompiler = webpack(getWebsiteWebpackConfig({
+      projectOptions,
+      production: true,
+      outputPath
+    }));
+
+    await new Promise((resolve, reject) => {
```
JoMarton: same as above

packages/website/src/server/index.ts
danielstro: I mean that instead of giving the type inline create an interface for it:
`interface bla { outputPath: string;  production: boolean;  projectOptions: IProjectOptions; }`
and then use it as the type for getWebsiteWebpackConfig and getMetadataWebpackConfig

JoMarton:
Please also update the README

alisey:
Will rename to showcase and update the readme in a separate PR.

PR #99, opened 2018-08-15
Linker - Adding get linked schema method

danielstro:
1. Added a get method that takes file path and export name and returns a linked schema (for the website).
2. Changed the CLI to allow getting linked schema (or regular one). At the moment this is only for testing and will be removed or changed in the future since we don't need a CLI command like that.
3. Solved circular types issue in schema extract so that we can run this on the mock repo. There are two parts, one is for the general handlers and one is specific for serializeType (since it has its own loop). serializeType also needs memoization and we should consider expanding that to the rest of the types.
4. Fixed some small things in linker so that it can run properly on the mock repo.
5. Added a test (currently failing) for handling intersection between interface and type. We need to this to properly handle react props.


packages/schema-extract/src/file-linker.ts
```diff
@@ -47,7 +48,10 @@ export class SchemaLinker {
                 refEntity = importSchema.definitions![entityType];
             }
         }
-        return refEntity ? refEntity : null;
+        if (refEntity && isRef(refEntity)) {
```
danielstro: Not TDD. I need to add a test for this but I am not sure how to recreate this issue

packages/schema-extract/src/file-linker.ts
```diff
@@ -267,20 +266,20 @@ export class SchemaLinker {
         if (entity.hasOwnProperty('constructor')) {
             res.constructor = Object.assign({}, entity.constructor);
         } else if (refEntity.hasOwnProperty('constructor')) {
-            res.constructor = Object.assign({}, refEntity.constructor);
+            res.constructor = Object.assign({}, (refEntity as ClassSchema).constructor);
```
danielstro: Although this is old code, I don't think we need Object.assign at all.

PR #103, opened 2018-08-23
WIP: Build component snapshots

Zombiefruit:


packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+const dotenv = require('dotenv');
+const {JSDOM} = require('jsdom');
+const {makeVisualGridClient, initConfig} = require('@applitools/visual-grid-client');
+const domNodesToCdt = require('@applitools/visual-grid-client/src/browser-util/domNodesToCdt');
+const {TestFailedError, TestResults} = require('@applitools/eyes.sdk.core');
+
+dotenv.config();
+
+function getGridClientConfig() {
+  const projectName = 'kieran-grid-test';
```
alisey: Let's take this from the `package.json` as we did before. And create a task on leankit for github integration, which will allow us to automatically set the `branchName`.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+  const branchName = projectName + '/master';
+
+  return {
+    appName: projectName,
+    branchName,
+    batchName: projectName,
+    browser: {
+      name: 'chrome',
+      width: 400,
```
alisey: 400 by 300 might be too small for some components, let's use 800 by 600.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+    resources[url] = {
+      url,
+      type: 'text/css',
+      value: fs.readFileSync(path.join(resourceDir, cssFilename))
+    };
+  }
+  return resources;
+}
+
+function getStatusIcon(result: any) {
```
alisey: `result: string`, and there's a lot of Anies below for which we can also specify the type.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+    sizeMode: 'viewport',
+    resourceContents: resources
+  });
+
+  return close();
+}
+
+export async function runEyes() {
+  const resourceDir   = path.join(process.cwd(), 'dist');
+  const cssFilenames  = glob.sync('*.css', {cwd: resourceDir});
```
alisey: This grabs all CSS files from `dist`, but what if there are some other files that weren't created by our tool? I think we shouldn't even put our CSS and HTML files into dist, because we don't plan to publish them to npm. Ideally we should use a temp folder. There's a `tmp` package on npm that does this.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+  const resources = getStaticResources(cssFilenames, resourceDir);
+  const gridClient = makeVisualGridClient(initConfig());
+
+  const testCount = htmlFilenames.length;
+  let resultCount = 0;
+  const resultPromises = [];
+
+  for (const htmlFilename of htmlFilenames) {
+    // tslint:disable-next-line:no-console
+    console.log(`Submitting ${htmlFilename}`);
```
alisey: I don't think we need this, even in verbose mode it wouldn't be that useful.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+        console.log(res);
+      }
+    });
+
+    resultPromises.push(result);
+  }
+
+  const results = await Promise.all(resultPromises);
+  // tslint:disable-next-line:no-console
+  console.log(results.map(getStatusIcon).join(' '));
```
alisey: We should just print the number of passing tests, new tests, failing tests, and how much time it took to run the batch.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  process.exit(0);
```
alisey: Should exit with status=1 if one of the tests fails.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+    url: 'http://localhost/html/index.html',
+    cdt: domNodesToCdt(dom),
+    sizeMode: 'viewport',
+    resourceContents: resources
+  });
+
+  return close();
+}
+
+export async function runEyes() {
```
alisey: Take a look at the `eyesTest` function we used to have, it checks the API key, gets the project name, exists with the correct status.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  for (const htmlFilename of htmlFilenames) {
+    // tslint:disable-next-line:no-console
+    console.log(`Submitting ${htmlFilename}`);
+    const html = fs.readFileSync(path.join(resourceDir, htmlFilename), 'utf-8');
+
+    const result = getTestResult(runTest(
+      gridClient,
+      config,
+      htmlFilename,
```
alisey: What is the format of the filename, does it contain the component name, style name, and simulation name?

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,172 @@
...
+      type: 'text/css',
+      value: fs.readFileSync(path.join(resourceDir, cssFilename))
+    };
+  }
+  return resources;
+}
+
+function logEyesResult({name, isNew, isModified, url, isError, error}: IResult) {
+  const fileSections = name.match(fileNameRegex);
+  const componentName = fileSections![1];
```
alisey: Create functions for generating / parsing the filename, and define them next to each other so it's easier to keep them in sync.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,172 @@
...
+  const simName = fileSections![3];
+  const variantName = fileSections![5];
+
+  const formattedUrl = isModified ? `${chalk.cyan('URL')}: ${chalk.underline(url)}` : '';
+  const status = isModified ? chalk.red('👎  FAIL') :
```
alisey: "Fail" and "error" are basically synonyms. I wouldn't know which one refers to "difference detected" and which one to "something went wrong while running the test". We could use terminology from Eyes: "match", "mismatch", "new". Or from git: "unmodified", "modified", "new", or anything else as long as different statuses are descriptive and clearly differentiated.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,172 @@
...
+               isNew ? chalk.yellow('👌  NEW') :
```
alisey: Yellow should be reserved for warnings imo. New is green.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env node
+
+const {TestFailedError, TestResults} = require('@applitools/eyes.sdk.core');
```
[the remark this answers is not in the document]
Zombiefruit: I don't think so, since typescript complains about the lack of typings, even if I do `import * as ...`

README.md
```diff
@@ -69,6 +70,22 @@ myComponentMetadata.addSim({
         items:['🧒 ', '👶 ', '🐊 ']
     }
 });
+
+// If you want to use the "snap" tool, this method must be called. Currently, the snap
+// tool relies on Stylable
+myComponentMetadata.exportedFrom({
+  path: 'src/my-comp/my-comp',                          // the path to your component, relative to the root, and without file extension
+  importName: 'MyComp',                                 // the name under which you export your component
+  baseStylePath: 'src/my-comp/my-comp/my-comp.st.css',  // the path to the base stylesheet for the component (as opposed to themes)
```
alisey: One too many `my-comp`'s

README.md
```diff
@@ -69,6 +70,22 @@ myComponentMetadata.addSim({
...
+  displayName: 'Composite'                              // the name you want to refer to the component with. Does not need to match importName
```
alisey: I'm wondering why it would different from `importName` or `displayName` defined in the component file.

README.md
```diff
@@ -69,6 +70,22 @@ myComponentMetadata.addSim({
         items:['🧒 ', '👶 ', '🐊 ']
     }
 });
+
+// If you want to use the "snap" tool, this method must be called. Currently, the snap
+// tool relies on Stylable
+myComponentMetadata.exportedFrom({
+  path: 'src/my-comp/my-comp',                          // the path to your component, relative to the root, and without file extension
+  importName: 'MyComp',                                 // the name under which you export your component
```
alisey: "name under which you export your component", so `exportName`?

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,168 @@
...
+const {makeVisualGridClient, initConfig} = require('@applitools/visual-grid-client');
+const domNodesToCdt = require('@applitools/visual-grid-client/src/browser-util/domNodesToCdt');
+import * as path from 'path';
+import * as fs from 'fs';
+import glob from 'glob';
+import chalk from 'chalk';
+import {JSDOM} from 'jsdom';
+import {consoleLog} from '@ui-autotools/utils';
+import {parseFilename} from '../generate-snapshots/filename-utils';
```
alisey: I'd call it `parseSnapshotFilename`. `parseFilename` sounds like something that parses a filename into basename and extension.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,168 @@
...
+      url,
+      type: 'text/css',
+      value: fs.readFileSync(path.join(resourceDir, cssFilename))
+    };
+  }
+  return resources;
+}
+
+function logEyesResult({name, status, url, error}: IResult) {
+  const {compName, simName, styleName} = parseFilename(name, '.snapshot.html');
```
alisey: I think that `parseFilename` should be aware that filenames end with `.snapshot.html` instead of receiving that as an argument.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,168 @@
...
+// await on a bunch of them in parallel using Promise.all, and a single rejected
+// promise would reject the entire batch.
+function getTestResult(testName: string, testResult: Promise<any>): Promise<IResult> {
+  return (
+    testResult
+    .catch((err: any) => err)
+    .then((res: any) => res instanceof TestFailedError ? res.getTestResults() : res)
+    .then((res: any) => Array.isArray(res) && res[0] instanceof TestResults ? res[0] : res)
+    .then((res: any): IResult => {
+      const url = res.getUrl();
```
alisey: This method is only available on instances of `TestResults` and should only be used in the `if` below.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,168 @@
...
+  const config = getGridClientConfig(projectPath);
+  const resources = getStaticResources(cssFilenames, directory);
+  const gridClient = makeVisualGridClient(initConfig());
+
+  const resultPromises = [];
+  consoleLog('Sending snapshots to Applitools...');
+
+  for (const htmlFilename of htmlFilenames) {
+    const html = fs.readFileSync(path.join(directory, htmlFilename), 'utf-8');
+    const {testName, testResult} = await runTest(
```
alisey: Here we're passing htmlFilename as `testName` to `runTest` function, which returns it back to us, and we assign it to `testName`.
1) There's no reason to return testName from runTest, because we already have it.
2) Test name should not be the same as the filename. We're sending it to Applitools and that's the string users will see as the title of the test in the dashboard.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,170 @@
+const {TestFailedError, TestResults} = require('@applitools/eyes.sdk.core');
+const {makeVisualGridClient, initConfig} = require('@applitools/visual-grid-client');
+const domNodesToCdt = require('@applitools/visual-grid-client/src/browser-util/domNodesToCdt');
+import * as path from 'path';
```
alisey:
```
import path from 'path';
import fs from 'fs';
```
Should work for these.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,170 @@
...
+import * as fs from 'fs';
+import glob from 'glob';
+import chalk from 'chalk';
+import {JSDOM} from 'jsdom';
+import {consoleLog} from '@ui-autotools/utils';
+import {parseSnapshotFilename} from '../generate-snapshots/filename-utils';
+
+interface IResult {
```
alisey: `ITestResult` or `IEyesResult` would be more descriptive.

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,170 @@
...
+      config,
+      testName,
+      html,
+      resources
+    ));
+
+    result.then((res: IResult) => {
+      if (res.status === 'error' || res.status === 'modified') {
+        process.exitCode = 1;
+        logEyesResult(res);
```
alisey: Now I'm thinking that we don't really need to pass testName to `getTestResults` only to receive it back again, and instead we can pass it here: `logEyesResult(testName, res)`

packages/eyes/src/generate-snapshots/generate-mapping.ts
```diff
@@ -0,0 +1,15 @@
+import * as path from 'path';
+import {IRegistry} from '@ui-autotools/registry';
+
+export function generateMapping(Registry: IRegistry) {
+  const mapping: any = {};
+
+  Registry.metadata.components.forEach((componentMetadata) => {
+    if (componentMetadata.compInfo.exportName) {
+      // We map styles to their components here (i.e. blah/blah/style1.st.css -> blah/blah/myComp.tsx)
+      mapping[path.join(process.cwd(), componentMetadata.compInfo.baseStylePath)] = path.join(process.cwd(), componentMetadata.compInfo.path);
```
alisey: Should receive path as an argument instead of using `process.cwd()`. The only place we use process.cwd() should be in the cli package.

packages/eyes/src/generate-snapshots/generate-mapping.ts
```diff
@@ -0,0 +1,15 @@
+import * as path from 'path';
```
alisey: import path from 'path';

packages/eyes/src/generate-snapshots/filter-logic.ts
```diff
@@ -0,0 +1,32 @@
+export function generateFilteringLogic(mapping: any): (stylableModule: any) => any {
```
alisey: It would be great to have a comment here explaining what this is all about. Maybe some of the any's could be replaced with types as well?

packages/eyes/src/generate-snapshots/generate-mapping.ts
```diff
@@ -0,0 +1,15 @@
...
+import {IRegistry} from '@ui-autotools/registry';
+
+export function generateMapping(Registry: IRegistry) {
+  const mapping: any = {};
```
alisey: Should be string to string.

packages/eyes/src/generate-snapshots/generate-mapping.ts
```diff
@@ -0,0 +1,15 @@
+import * as path from 'path';
+import {IRegistry} from '@ui-autotools/registry';
+
+export function generateMapping(Registry: IRegistry) {
```
alisey: How about `mapStylesToComponents` or `generateMappingOfStylesToComponents`?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
+import * as path from 'path';
```
alisey: import path from 'path';

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+const StylableWebpackPlugin = require('@stylable/webpack-plugin');
+const config = require(path.join(process.cwd(), './.autotools/webpack.config.js'));
+import * as webpack from 'webpack';
```
alisey: import webpack from 'webpack';

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
+import * as path from 'path';
+const StylableWebpackPlugin = require('@stylable/webpack-plugin');
+const config = require(path.join(process.cwd(), './.autotools/webpack.config.js'));
```
alisey: Should not use `process.cwd()`

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as webpack from 'webpack';
+import * as glob from 'glob';
```
alisey: import glob from 'glob';

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+      new HTMLSnapshotPlugin({
+        render(compiledFile, sourceFile) {
+          let compMetadata: IComponentMetadata<any> | undefined;
+          const mapIter = Registry.metadata.components.entries();
+          while (!compMetadata) {
+            // We only have to do this because we currently map the component definitions to their metadata,
+            // not the names. So we have no way to get component metadata by name. And at this point in the build
+            // process, the component definition returned by webpack has been modified from the original, so we can't
+            // get the metadata with it
+            const currentIteration = mapIter.next();
```
alisey: Why not a `for ... of` loop or `.find`?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+
+importMeta();
+
+const mapping = generateMapping(Registry);
+const filteringLogic = generateFilteringLogic(mapping);
+
+const fileNameRegex = /(?:.\/.autotools\/)(.+)/; // Match the file name
+
+async function buildSingleFile(file: string, directory: string) {
+  const entryName = fileNameRegex.exec(file)![1];
```
alisey: Use `match` instead of `exec` because it's stateless.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+          </body>
+          </html>`;
+          return dedent(template);
+        },
+        getLogicModule: filteringLogic
+    })
+    ]
+  };
+
+  const mergedConfig = {...config, ...snapshotConfig};
```
alisey: If the config has plugins will this discard them?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+  const compiler = webpack.default(mergedConfig);
```
alisey: Change the import for `webpack` and you'll be able to use `webpack(mergedConfig)` here.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+              break;
+            }
+          }
+
+          if (!compMetadata) {
+            throw new Error(`Could not find component metadata for ${compiledFile.default.name}`);
+          }
+
+          const {simIndex} = parseSnapshotFilename(sourceFile.id, '.ts');
+          const props = compMetadata.simulations[simIndex - 1].props;
```
alisey: Why `-1`?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+          const link = `<link rel="stylesheet" type="text/css" href="${entryName}.css">`;
```
alisey: `cssLink`

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+      path: directory
+    },
+    plugins: [
+      new StylableWebpackPlugin({
+        outputCSS: true,
+        path: directory,
+        filename: '[name].css'
+      }),
+      new HTMLSnapshotPlugin({
+        render(compiledFile, sourceFile) {
```
alisey: Define this as a top-level function to reduce indentation and to make it easier to see which parts of webpack config we adjust.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
+export const generateSnapshots = async (directory: string) => {
+  const files = glob.sync('./.autotools/**.ts'); // Tool must be run in the root
```
alisey: Pass the project root as an argument.

packages/eyes/src/index.ts
```diff
@@ -1,148 +1,21 @@
-const {Eyes} = require('eyes.images');
+import {buildBaseFiles} from './generate-snapshots/build-base-files';
+import {generateSnapshots} from './generate-snapshots/generate-snapshots';
+import {runEyes} from './eyes-test/eyes';
+import * as tmp from 'tmp';
```
alisey: import tmp from 'tmp';

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,108 @@
...
+import {createElement} from 'react';
+import {renderToStaticMarkup} from 'react-dom/server';
+import {generateFilteringLogic} from './filter-logic';
+import {generateMapping} from './generate-mapping';
+import Registry, {importMeta, getCompName, IComponentMetadata} from '@ui-autotools/registry';
+import {consoleLog} from '@ui-autotools/utils';
+import {dedent} from './dedent';
+import {parseSnapshotFilename} from './filename-utils';
+
+importMeta();
```
alisey: Move these three lines to `generateSnapshots`. Since this is not a standalone script importing it shouldn't do anything.

packages/mock-repo/.autotools/Composite@1@compositeSim@1@style1.ts
```diff
@@ -0,0 +1,3 @@
+import style from '../src/composite/variant1.st.css';
```
alisey: These files shouldn't exist, right?

packages/mock-repo/.autotools/webpack.config.js
```diff
@@ -1,4 +1,6 @@
-const packagePath = __dirname;
+const path = require('path');
+const packagePath = path.join(__dirname, '..');
```
alisey: `path.resolve`

packages/mock-repo/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@ui-autotools/registry": "^1.0.7",
     "@ui-autotools/scripts": "^1.0.19",
+    "@ui-autotools/utils": "^1.0.6",
```
alisey: Which of utils does it use?

packages/mock-repo/src/composite/composite.tsx
```diff
@@ -1,11 +1,13 @@
...
+import style from './child.st.css';
 
 export interface IProps {
   text?: React.ReactNode;
+  className?: string;
 }
 
 export const Composite: React.SFC<IProps> = (props) =>
-  <div><ChildComp text={props.text} /></div>;
+  <div  {...style('root', {}, props)} className={props.className}><ChildComp text={props.text} /></div>;
```
alisey: You shouldn't need to add `className` explicitly, `...style()` should do that for you.

packages/mock-repo/test/eyes.spec.ts
```diff
@@ -0,0 +1,44 @@
+const {makeVisualGridClient, initConfig} = require('@applitools/visual-grid-client');
+const domNodesToCdt = require('@applitools/visual-grid-client/src/browser-util/domNodesToCdt');
+import * as path from 'path';
```
alisey: `import path from 'path';` – look for other `import * as` and replace them too.

packages/mock-repo/test/eyes.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    }));
+  });
+  afterEach(() => closePromises.push(close()));
+
+  it('should work', () => {
+    checkWindow({
+      tag: 'first test',
+      url: 'http://localhost/index.html',
+      cdt: domNodesToCdt(
+        new JSDOM(fs.readFileSync(path.join(__dirname, '../dist/snapshots/variant1.snapshot.html'), 'utf-8')).window
```
alisey: Why is it in the dist?

packages/mock-repo/test/eyes.spec.ts
```diff
@@ -0,0 +1,44 @@
+const {makeVisualGridClient, initConfig} = require('@applitools/visual-grid-client');
```
alisey: Let's not test the grid client in separation, it's a third-party package that should have its own tests, and it's not our job to test third-party code.

If we want to test our code then we need to replace this library with a mock. Our tests should not submit anything to Applitools.

packages/registry/src/registry/component-metadata.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react';
-import {IComponentMetadata, ISimulation, IStyleMetadata} from './types';
+import {IComponentMetadata, ISimulation, IStyleMetadata, ICompInfo} from './types';
 import {isValidSimulationTitle} from '../utils';
 
 export class ComponentMetadata<Props> implements IComponentMetadata<Props> {
   public simulations: Array<ISimulation<Props>> = []; // Initialize with "empty" simulation
   public styles: Map<any, IStyleMetadata> = new Map<any, IStyleMetadata>();
   public reactStrictModeCompliant: boolean = true;
+  public compInfo: ICompInfo = {path: '', exportName: '', baseStylePath: ''};
```
alisey: Component metadata contains information about the component, now we have information a as subset of information. I would suggest either to remove this nesting and add `exportPath`, `exportName`, and `baseStylePath` directly on the metadata object, or to rename `compInfo` to something like `exportInfo`.

packages/registry/src/utils/import-meta-files.ts
```diff
@@ -0,0 +1,22 @@
+import 'typescript-support';
+import * as glob from 'glob';
+import {registerRequireHooks} from '@ui-autotools/utils';
+
+registerRequireHooks();
+
+const importMeta = (filePattern?: string) => {
```
alisey: Should receive base path.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
Zombiefruit: Changed

packages/mock-repo/package.json
Zombiefruit: consoleLog

alisey: Don't see it anywhere.

Zombiefruit: Whoops, my bad. Thought it was in Eyes. Yeah, deleted this dep

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+    fs.mkdirSync(autotoolsFolder);
+  }
+
+  const stylePathPrefix = '../';
+  const compPathPrefix = '../';
+
+  Registry.metadata.components.forEach((componentMetadata) => {
+    const numberOfSims = componentMetadata.simulations.length;
+    const styles = componentMetadata.styles;
+    const compPath = compPathPrefix + componentMetadata.path;
```
alisey: `path.join(projectPath, componentMetadata.path)`

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+    const compName = componentMetadata.exportName;
+
+    if (compName) {
+      for (let i = 0; i < numberOfSims; i++) {
```
alisey: Call it `simIndex`

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+        const simulationName = componentMetadata.simulations[i].title;
+        if (styles.size) {
+          let styleIndex = 1;
+          styles.forEach((style) => {
+            const stylePath = stylePathPrefix + style.path;
```
alisey: `path.join(projectPath, style.path)`

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+    const numberOfSims = componentMetadata.simulations.length;
+    const styles = componentMetadata.styles;
+    const compPath = compPathPrefix + componentMetadata.path;
+    const compName = componentMetadata.exportName;
+
+    if (compName) {
+      for (let i = 0; i < numberOfSims; i++) {
+        const simulationName = componentMetadata.simulations[i].title;
+        if (styles.size) {
+          let styleIndex = 1;
```
alisey: Add a comment explaining that we only want to render the base style if there are no variant styles, and that index 0 is reserved for the base style, and that's why we start at 1.

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+          styles.forEach((style) => {
```
alisey: Instead of defining styleIndex above and incrementing in manually you could do:
```
styles.forEach((style, styleIndex) => {
    generateFilename(..., styleIndex + 1);
})
```

packages/eyes/src/eyes-test/eyes.ts
```diff
@@ -0,0 +1,169 @@
...
+      value: fs.readFileSync(path.join(resourceDir, cssFilename))
+    };
+  }
+  return resources;
+}
+
+function formatName(filename: string) {
+  const {compName, simName, styleName} = parseSnapshotFilename(filename, '.snapshot.html');
+
+  return `${compName}: ${simName}. Style: ${styleName}.`;
```
alisey: If we're using the base style we need to return simply `${compName}: ${simName}.`

packages/eyes/src/generate-snapshots/filter-logic.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * This function exports a function, which is done to give the inner function (filterLogicModule) access
```
alisey: It's clear without comments that this function binds the first argument to a different function. The question for me here is what does this code do, `generateFilteringLogic` is too abstract.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+import {consoleLog} from '@ui-autotools/utils';
+import {dedent} from './dedent';
+import {parseSnapshotFilename} from './filename-utils';
+
+const fileNameRegex = /(?:.\/.autotools\/)(.+)/; // Match the file name
+
+async function buildSingleFile(file: string, directory: string, filteringLogic: (stylableModule: any) => any, config: any) {
+  const entryName = file.match(fileNameRegex)![1];
+
+  function render(compiledFile: any, sourceFile: any) {
```
alisey: Is there a reason why this function needs to be defined inside of `buildSingleFile` instead of the top level?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+    let compMetadata: IComponentMetadata<any> | undefined;
+
+    // We only have to do this because we currently map the component definitions to their metadata,
+    // not the names. So we have no way to get component metadata by name. And at this point in the build
+    // process, the component definition returned by webpack has been modified from the original, so we can't
+    // get the metadata with it
+    for (const component of Registry.metadata.components) {
```
alisey: Define `findComponentByName` as a separate function.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+        filename: '[name].css'
+      }),
+      new HTMLSnapshotPlugin({
+        render,
+        getLogicModule: filteringLogic
+    })
+    ]
+  };
+
+  const mergedConfig = {...config, ...snapshotConfig};
```
alisey: Add a `TODO` comment here explaining that we should use the project's Stylable instead of replacing it with our own.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
+export const generateSnapshots = async (processDir: string, directory: string) => {
```
alisey: It's not clear what is the `processDir` and what is `directory`.

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+  importMeta();
```
alisey: `buildBaseFiles` also imports meta, does it mean we're importing it twice?

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+  const config = require(path.join(processDir, './.autotools/webpack.config.js'));
```
alisey: Call it `webpackConfig`

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+  const mapping = mapSylesToComponents(Registry, processDir);
+  const filteringLogic = generateFilteringLogic(mapping);
+  const files = glob.sync('./.autotools/**.ts', {cwd: processDir}); // Tool must be run in the root
```
alisey: In the root of what and why does it matter here? Why is it looking for files in `.autotools`? That folder is not supposed to contain any generated files.

packages/eyes/src/generate-snapshots/map-styles-to-components.ts
```diff
@@ -0,0 +1,15 @@
+import path from 'path';
+import {IRegistry} from '@ui-autotools/registry';
+
+export function mapSylesToComponents(Registry: IRegistry, currentDirectory: string): {[stylePath: string]: string} {
```
alisey: Rename `currentDirectory` to `projectDirectory`.

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+            const stylePath = stylePathPrefix + style.path;
+            const filename = generateFilename(compName, simulationName, i, style.name, styleIndex);
+            const data = generateData(compName, compPath, stylePath);
+            fs.writeFileSync(filename, data);
```
alisey: Which folder does it write to? Is it relative to the currentDir?

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const buildBaseFiles = (projectPath: string) => {
+  consoleLog('Building base files...');
+  consoleLog('Importing Metafiles...');
+  importMeta();
+
+  const autotoolsFolder = path.join(projectPath, '.autotools');
+
+  if (!fs.existsSync(autotoolsFolder)) {
+    fs.mkdirSync(autotoolsFolder);
```
alisey: What's the point of creating this directory, I don't see that we're using it later in this file.

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,47 @@
+import Registry, {importMeta} from '@ui-autotools/registry';
+import {consoleLog} from '@ui-autotools/utils';
+import fs from 'fs';
+import path from 'path';
+import {generateFilename, generateData} from './filename-utils';
+
+export const buildBaseFiles = (projectPath: string) => {
```
alisey: Should receive registry as an argument instead of importing meta.

packages/eyes/src/generate-snapshots/build-base-files.ts
Zombiefruit: Using it now

packages/eyes/src/generate-snapshots/generate-snapshots.ts
Zombiefruit: Ah yeah, it needs to have `entryName` in its context

packages/eyes/src/generate-snapshots/generate-snapshots.ts
Zombiefruit: Importing once now

packages/eyes/src/generate-snapshots/map-styles-to-components.ts
```diff
@@ -0,0 +1,15 @@
+import path from 'path';
+import {IRegistry} from '@ui-autotools/registry';
+
+export function mapSylesToComponents(Registry: IRegistry, projectDirectory: string): {[stylePath: string]: string} {
+  const mapping: {[stylePath: string]: string} = {};
+
+  Registry.metadata.components.forEach((componentMetadata) => {
+    if (componentMetadata.exportName) {
+      // We map styles to their components here (i.e. blah/blah/style1.st.css -> blah/blah/myComp.tsx)
```
alisey: This comment makes sense as the description of the entire function. Also it shows relative paths in the example, which is slightly confusing because we're mapping absolute paths.

```
Map paths of base stylesheets to paths of respective components, e.g.:
/my-proj/src/checkbox.st.css -> /my-proj/src/checkbox.tsx
```

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
+export const generateSnapshots = async (processDir: string, tempDirectory: string, Registry: IRegistry) => {
```
alisey: processDir -> projectDir

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+  for (const component of Registry.metadata.components) {
+      const metadata = component[1];
+      if (getCompName(metadata.component) === name) {
+        return metadata;
+      }
+    }
+}
+
+async function buildSingleFile(fileName: string, filePath: string, directory: string, filteringLogic: (stylableModule: any) => any, config: any, Registry: IRegistry) {
+  function render(compiledFile: any, sourceFile: any) {
```
alisey: Move `render` out of this function and provide `fileName` either by binding it:
```
new HTMLSnapshotPlugin({
   render: render.bind(null, fileName)
})
```
or by defining an anonymous function:
```
new HTMLSnapshotPlugin({
  render: (...args) => render(fileName, ..args)
})
```
Do the same for `filteringLogic`

packages/eyes/src/generate-snapshots/generate-snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+import {createElement} from 'react';
+import {renderToStaticMarkup} from 'react-dom/server';
+import {generateFilteringLogic} from './filter-logic';
+import {mapSylesToComponents} from './map-styles-to-components';
+import {getCompName, IComponentMetadata, IRegistry} from '@ui-autotools/registry';
+import {consoleLog} from '@ui-autotools/utils';
+import {dedent} from './dedent';
+import {parseSnapshotFilename} from './filename-utils';
+
+function findComponentByName(name: string, Registry: IRegistry): IComponentMetadata<any> | void {
```
alisey: Indentation is messed up in this function.

packages/eyes/src/generate-snapshots/filter-logic.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * This method is used by the HTMLSnapshotPlugin to determine which component logic file to link to a specific style
```
alisey: 😀

packages/eyes/src/generate-snapshots/filename-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface IFileParts {
+  base: string;
+  compName: string;
+  simIndex: number;
+  simName: string;
+  styleIndex?: number;
+  styleName?: string;
+}
+
+export function generateFilename(componentName: string, simName: string, simIndex: number, styleName?: string): string {
```
alisey: `generateSnapshotFilename`

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,45 @@
...
+    const styles = componentMetadata.styles;
+    const compPath =  path.join(compPathPrefix, componentMetadata.path);
+    const compName = componentMetadata.exportName;
+
+    if (compName) {
+      for (let i = 0; i < simIndex; i++) {
+        const simulationName = componentMetadata.simulations[i].title;
+        if (styles.size) {
+          styles.forEach((style) => {
+            const stylePath = path.join(stylePathPrefix, style.path);
```
alisey: Use `projectPath`. `stylePath` is relative to the current folder, and assumes that the current folder is the project folder.

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,45 @@
...
+    fs.mkdirSync(autotoolsFolder);
+  }
+
+  const stylePathPrefix = '../../'; // We're two folders deep in .autotools
+  const compPathPrefix = '../../';
+
+  Registry.metadata.components.forEach((componentMetadata) => {
+    const simIndex = componentMetadata.simulations.length;
+    const styles = componentMetadata.styles;
+    const compPath =  path.join(compPathPrefix, componentMetadata.path);
```
alisey: Use `projectPath`

packages/eyes/src/generate-snapshots/build-base-files.ts
```diff
@@ -0,0 +1,45 @@
+import {consoleLog} from '@ui-autotools/utils';
+import fs from 'fs';
+import path from 'path';
+import {generateFilename, generateData} from './filename-utils';
+import { IRegistry } from '@ui-autotools/registry';
+
+export const buildBaseFiles = (projectPath: string, Registry: IRegistry) => {
+  consoleLog('Building base files...');
+  consoleLog('Importing Metafiles...');
```
alisey: It's not importing meta files.

PR #105, opened 2018-08-27
Showcase: a basic props table

alisey:
Very alpha stage, throws an exception for most of the components, but it's a start.

Checkbox in `wix-ui-core`:

image


alisey:
Checkbox in wix-ui-core:

image

PR #106, opened 2018-08-28
Linker - Create program optimization

danielstro:
Instead of creating a program multiple times for each file, we create one once for all of the files.

### DIFF
--- a/packages/schema-extract/src/extract-schema.ts
+++ b/packages/schema-extract/src/extract-schema.ts
@@ -6,13 +6,15 @@ import {transform} from './file-transformer';
 import {createHost} from './isomorphc-typescript-host';
 import {SchemaLinker} from './file-linker';
 
+let program: typescript.Program;
+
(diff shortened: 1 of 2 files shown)


packages/schema-extract/src/extract-schema.ts
```diff
@@ -21,21 +23,27 @@ export function* extractSchema(basePath: string, filesGlob: string) {
   }
 }
 
+export function createLinkerProgram(files: string[]) {
+  program = typescript.createProgram(files, {});
+}
+
 export function getSchema(filePath: string, exportName: string) {
-  const program = typescript.createProgram([filePath], {});
+  if (!program) {
```
alisey: Instead of having the program as a global state I'd prefer it to be an argument of getSchema, or maybe even better to have a class:
```
class LinkerSomething {
   program = null;
   constructor(files) {
     this.program = createLinkerProgram();
   }
   getSchema(...) {
     // Use this.program
   }
}
```

PR #130, opened 2018-09-05
Showcase: style dropdown

alisey:


packages/showcase/src/client/simulation.tsx
```diff
@@ -1,22 +1,67 @@
...
-const simulationTitle = url.searchParams.get('simulation');
+function findComponent(compName: string) {
+  const comps = Array.from(Registry.metadata.components.keys());
+  return comps.find((c) => getCompName(c) === compName);
+}
 
-const Comp = Array.from(Registry.metadata.components.keys()).find((c) =>
-  getCompName(c) === componentName
-);
+function findSimulation<T>(compMeta: IComponentMetadata<T>, simName: string) {
```
danielstro: Does findSimulation need to be generic? You are not calling it with an argument when you use it. 
You can use IComponentMetadata<any> instead

packages/showcase/src/client/simulation.tsx
```diff
@@ -1,22 +1,67 @@
...
+  return compMeta.simulations.find(({title}) => title === simName);
+}
 
-const compMeta = Comp && Registry.metadata.components.get(Comp);
-const sim = compMeta && compMeta.simulations.find(({title}) =>
-  title === simulationTitle
-);
+function findStyle<T>(compMeta: IComponentMetadata<T>, styleName: string) {
```
danielstro: same

packages/showcase/src/client/simulation.tsx
```diff
@@ -1,22 +1,67 @@
...
+  for (const [style, styleMeta] of compMeta.styles) {
```
danielstro: compMeta.styles is map, why not use .get?

PR #138, opened 2018-10-04
work on showcase docs

nadavwix:


packages/showcase/README.md
```diff
@@ -2,6 +2,114 @@
...
+you can categorize components using the meta-data-repo:
+
+```ts
+import Registry from '@ui-autotools/registry';
+import {Composite} from './composite';
+import style1 from './variant1.st.css';
+
+const metadata = Registry.getComponentMetadata(Composite);
+
+metadata.categorize('components/inputs/date/date-input');
```
JoMarton: I'm worried about when we'd want to do some tree restructuring like adding additional level or even just renaming of directories. Maybe if we just reference to a parent directory it will be simpler to manage in site settings

packages/showcase/README.md
```diff
@@ -2,6 +2,114 @@
...
+
+```
+
+### categorization using a comment:
+
+categorizing non component code assets is done useing a js-doc comment:
+
+```ts
+/**
+* @category components/inputs/date/date-formater
```
JoMarton: given my comment above this can be something like `@categoryParent date` and `@name date-formater`

packages/showcase/README.md
```diff
@@ -2,6 +2,114 @@
...
+
+notice that if that page does not exist, a new one is added.
+
+
+
+```ts
+import * as React from 'react';
+import Registry from '@ui-autotools/registry';
+
+const metadata = Registry.getPage('components/inputs/date/date-input');
```
JoMarton: same comment as for `.categorize`

AviVahl:
irrelevant.

PR #146, opened 2018-10-11
State override

Zombiefruit:
Add the ability to override state in simulations.


packages/registry/src/utils/get-comp-with-state.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
JoMarton: Please add detailed comments for this file

packages/mock-repo/src/composite/composite.tsx
```diff
@@ -7,7 +7,17 @@ export interface IProps {
...
-  <div {...style('root', {}, props)}><ChildComp text={props.text} /></div>;
+export interface IState {
+  text?: React.ReactNode;
+}
+
+export class Composite extends React.Component<IProps, IState> {
+  public static displayName: string;
+  public state: IState = {text: ''};
+
+  public render() {
```
JoMarton: I think we need an example that's a bit more "realistic", like using `setState` in the comp

alisey: Something like a counter that increments on each click. Or a dropdown with `open` state.

packages/registry/src/utils/get-comp-with-state.tsx
```diff
@@ -3,43 +3,27 @@ import {ISimulation} from '../registry';
...
-      // Save the old state
-      const oldState = super.state;
-      // Replace the old state with the old state + the override state
-      super.state = {...oldState, ...simulation.state};
-    }
-
-    public render() {
-      // Just return the extended Component's render method
-      return super.render();
+      this.state = {...this.state, ...simulation.state};
```
alisey:
```
class Wrapped extends Comp {
  constructor() {
    ...
  }

  setState(state) {
    super.setState(getOverrideState(state));
  }
}
```

Not sure if this is important at this point, but `setState()` can also receive an updater function and a callback. And there's also `getDerivedStateFromProps` which can affect the state.

packages/registry/src/utils/get-comp-with-state.tsx
```diff
@@ -0,0 +1,29 @@
...
+    };
+  };
+
+  class Wrapped extends Comp {
+    constructor(props: any) {
+      super(props);
+      this.state = getOverrideState(this.state);
+    }
+
+    public setState(o: (args?: any[]) => any, callback?: (() => void) | undefined) {
```
alisey: Isn't `| undefined` redundant since the argument is optional?

packages/mock-repo/src/counter/counter.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+
+interface IState {
+  count: number;
+}
+
+export class Counter extends React.Component<{}, IState> {
+  public static displayName: string = 'Counter';
+
+  constructor(props: {}) {
```
alisey: Don't need constructor here
```
export class Counter {
  public state: IState = {count: 0};
}
```

packages/mock-repo/src/counter/counter.tsx
```diff
@@ -0,0 +1,26 @@
...
+    super(props);
+    this.state = { count: 0 };
+  }
+
+  public onClick = () => {
```
alisey: Should be private.

PR #152, opened 2018-10-16
Actually replace ts-loader with @ts-tools/webpack-loader

Zombiefruit:


packages/schema-extract/webpack.config.js
```diff
@@ -15,12 +15,11 @@ module.exports = {
...
             {
                 test: /\.tsx?$/,
-                loader: 'ts-loader',
+                loader: '@ts-tools/webpack-loader',
                 options: {
                     compilerOptions: {
                         declaration: false,
-                        declarationMap: false,
-                        module: 'esnext'
+                        declarationMap: false
```
alisey: We don't need the entire options object anymore since webpack-loader turns declarations off by default.

packages/showcase/src/server/index.ts
```diff
@@ -82,7 +82,7 @@ function getWebsiteWebpackConfig(
         {
           test: /\.tsx?$/,
           use: {
-            loader: 'ts-loader',
+            loader: '@ts-tools/webpack-loader',
```
alisey: And here as well we don't need `options`

PR #153, opened 2018-10-17
Refactor snap to use filesystem as a parameter

Zombiefruit:


packages/snap/src/index.ts
```diff
@@ -1,13 +1,13 @@
 import {buildBaseFiles} from './generate-snapshots/build-base-files';
 import {generateSnapshots} from './generate-snapshots/generate-snapshots';
 import {runEyes} from './snap-test/snap';
-import { createTempDirectory } from 'create-temp-directory';
+import {createTempDirectory} from 'create-temp-directory';
 import Registry from '@ui-autotools/registry';
 
-export async function eyesTest(projectPath: string) {
+export async function eyesTest(projectPath: string, fs: any) {
```
alisey: We could define our own interface that contains only relevant functions. Something like:
```
interface IFilesystem {
  writeFileSync: (file: string, data: string) => void;
  ...
}
```

packages/snap/src/snap-test/snap.ts
```diff
@@ -133,12 +132,12 @@ async function runTest(gridClient: any, gridClientConfig: any, testName: string,
   return close();
 }
 
-export async function runEyes(projectPath: string, tempDirectory: string) {
+export async function runEyes(projectPath: string, tempDirectory: string, fs: any) {
```
alisey: If glob always works with the real file system, then we're screwed.

packages/snap/src/generate-snapshots/build-base-files.ts
```diff
@@ -3,7 +3,12 @@ import {generateSnapshotFilename, generateData} from './filename-utils';
 import { IRegistry } from '@ui-autotools/registry';
 import path from 'path';
 
-export const buildBaseFiles = (projectPath: string, Registry: IRegistry, fs: any) => {
+export interface IFileInfo {
+  filename: string;
```
alisey: I'd rename `filename` to `basename` to make it more clear that it doesn't include the extension.

packages/snap/src/generate-snapshots/filename-utils.ts
```diff
@@ -1,7 +1,5 @@
-import {basename} from 'path';
-
 export interface IFileParts {
-  base: string;
+  file: string;
```
alisey: You don't seem to be using this field.

Zombiefruit: Probably Yagni... just wanted to leave it in in case

packages/snap/src/snap-test/snap.ts
```diff
@@ -133,26 +133,23 @@ async function runTest(gridClient: any, gridClientConfig: any, testName: string,
...
+  for (const file of files) {
+    const html = fs.readFileSync(path.join(tempDirectory, file.basename + '.snapshot.snapshot.html'), 'utf-8');
+    const testName = formatName(file.basename);
 
     const result = getTestResult(runTest(
       gridClient,
       config,
       testName,
-      html,
+      html as string,
```
alisey: Do we still need `as string`?

PR #154, opened 2018-10-18
Linker - intersection

danielstro:
Rewrote intersection handling (And changed some few things in class handling as a result).


packages/schema-extract/src/file-linker.ts
```diff
@@ -118,84 +117,55 @@ export class SchemaLinker {
...
+                }
+            }
+        }
+        for (const p in object2.properties) {
+            if (!res.properties.hasOwnProperty(p)) {
+                res.properties[p] = object2.properties[p];
+                if (object2.definedAt) {
+                    res.properties[p].definedAt = object2.definedAt;
+                }
+            } else {
```
alisey: If this can only happen when we have an invalid intersection type, then the question is should we even try to handle invalid type, or should we leave to typescript to check that the code we're working with doesn't have invalid types.

packages/schema-extract/src/file-linker.ts
```diff
@@ -255,26 +225,28 @@ export class SchemaLinker {
             properties: {},
             staticProperties: {}
         } as ClassSchema;
+        const paramsMap = new Map();
+        let isGeneric = false;
```
alisey:
```suggestion
        let isGeneric = refEntity.genericParams && entity.extends!.genericArguments;
        if (isGeneric) {
            refEntity.genericParams.forEach(...)
        }
```

packages/schema-extract/src/file-linker.ts
danielstro: Okay, I changed handleIntersection to throw if we reach a "never" situation so we don't need to check for it here.

PR #164, opened 2018-10-24
Schema Playground: add schema views

alisey:


packages/schema-playground/src/playground.tsx
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import { IFileSystem } from '@file-services/types';
 import { IBaseHost } from '@file-services/typescript';
 import { transform } from '@ui-autotools/schema-extract/esm/file-transformer';
+import {BaseView as BaseSchemaView, defaultSchemaViewRegistry} from '@ui-autotools/schema-views/src';
```
Zombiefruit: Is there a reason we're not exporting these from the root?

alisey: It's because Stylable doesn't support `build:watch`, and we want changes in `schema-views` to be reflected in the playground.

PR #166, opened 2018-10-28
Nadav/data literals

nadavwix:


packages/schema-playground/src/playground.tsx
```diff
@@ -64,7 +65,7 @@ export class Playground extends React.PureComponent<IPlaygroundProps, IPlaygroun
         const typeChecker = program && program.getTypeChecker();
         const sourceFile = program && program.getSourceFile(this.props.filePath);
         if (typeChecker && sourceFile) {
-            const moduleSchema = transform(typeChecker, sourceFile, this.props.filePath, '/');
+            const moduleSchema = transform(typeChecker, sourceFile, this.props.filePath, '/', path.posix);
```
danielstro: Playground uses fs but since IFileSystem doesn't have parse I didn't want to use it. Should we pass it with a different type / any?

PR #167, opened 2018-10-29
Added check for API key and option to skip test if lacking an api key

Zombiefruit:


packages/scripts/src/cli.ts
```diff
@@ -48,11 +48,12 @@ program
 .command('snap')
 .description('compare components to the expected appearance using Applitools Eyes')
 .option('-f, --files [pattern]', 'metadata file pattern')
+.option('-s, --skip-on-missing-key [boolean]', 'Grep file')
```
JoMarton: what does 'Grep file' mean?

JoMarton: what does 'Grep file' mean?

Zombiefruit: ...whoopsie

PR #171, opened 2018-11-04
[DO NOT MERGE!!]Schema-extract Readme

danielstro:


packages/schema-extract/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Usage
+
+Import the `transform` function from `@ui-autotools/schema-extract`. This function recives the following five arguments:
+* checker (`typescript.TypeChecker`) - a typescript checker connected to the desired files
+* sourceFile (`typescript.SourceFile`) - the typescript source of that we want to transform
+* modulePath (`string`) - The path to the module base directory
+* projectPath (`string`) - The name of the project the files are in (Will be removed in the future)
+* pathUtil (`IFileSystemPath`) - A path utility to be used to access directories and file. (You can use `path.posix`)
+
```
JoMarton: would be great to have a few input/output examples along with command

packages/schema-extract/README.md
```diff
@@ -0,0 +1,79 @@
+# Schema Extract
+
```
JoMarton: link to schema extract playground project

packages/schema-extract/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+Schema extract transforms code files into [JSON-Schema](https://json-schema.org/) schemas. It is composed of two parts, the **schema-extract** and **linker**.
+
```
JoMarton: Maybe hoist the basic linker usage, that utilizes the default ts transformer here. I would add an example to show input, command, and output

packages/schema-extract/README.md
```diff
@@ -0,0 +1,79 @@
+# Schema Extract
+
+
```
JoMarton: "Schema extraction is a necessary step for automatic documentation"

packages/schema-extract/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+    import {SchemaLinker} from '@ui-autotools/?????';
+    import {myExtractor} from './secretStuff';
+
+    function linkSchema(filePath, typeName) {
+        const customLinker = new SchemaLinker(myExtractor);
+        return linker.flatten(filePath, typeName);
+    }
+```
```
JoMarton: Backlog - what we know we don't support yet

packages/schema-extract/README.md
```diff
@@ -0,0 +1,148 @@
+# Schema Extract
+
+
+Schema extract transforms code files into [JSON-Schema](https://json-schema.org/) schemas. Schema extraction is a necessary step to create automated documentation for your code.
+
+It is composed of two parts, the **schema-extract** and **linker**.
```
JoMarton: schema-extract --> TS transformer

packages/schema-extract/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+## Playground
+The `schema-extract` package has a playground that transforms your code to a schema as you type it.
```
JoMarton: transforms to a schema AND a "view" from that schema

packages/schema-extract/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+Schema extract transforms code files into [JSON-Schema](https://json-schema.org/) schemas. Schema extraction is a necessary step to create automated documentation for your code.
+
+It is composed of two parts, the **schema-extract** and **linker**.
+
+
+## Playground
+The `schema-extract` package has a playground that transforms your code to a schema as you type it.  
+To use the playground simply install `ui-autotools` and go to `/packages/schema-playground` and run `yarn start`.
```
JoMarton: No need for instructions here. We'll add them to the playground readme

packages/schema-extract/README.md
```diff
@@ -0,0 +1,147 @@
...
+      ],
+      "$ref": "common/interface"
+    }
+  }
+}
+```
+
+## TS Transformer
+
+Transforms a typescript source file into JSON-Schema.
```
JoMarton: non linked schema

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
+# Schema Extract
+
+
+Schema extract transforms code files into [JSON-Schema](https://json-schema.org/) schemas. Schema extraction is a necessary step to create automated documentation for your code.
+
+It is composed of two parts, the **ts transformer** and **linker**.
+
+
+## Playground
+The `schema-extract` package has a playground that allows live editing of your code and transforms it to a schema and displays the results as you type it.
```
JoMarton: link to the playground project

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Usage
+
+Import the `createLinker` function from `@ui-autotools/schema-extract`. This function receives a string array of file paths and returns an initalized linker class.
+To get the linked schema of a specific file invoke `linker.flatten(fileName)`:
```
Zombiefruit: `linker.flatten` is shown with one parameter here (fileName) but then used later with two parameters (fileName and interface?)

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
...
+        'myProject/src/file-b.ts'
+    ];
+    const linker = createLinker(files);
+    const linkedSchema = linker.flatten(files[0], 'IAnimal');
+
+```
+
+## Example
+
+Take a look at the following interface:
```
Zombiefruit: Tacking onto the comment above, is the linker for interfaces or whole files?

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
...
+      ],
+      "$ref": "common/interface"
+    }
+  }
+}
+```
+
+## TS Transformer
+
+Transforms a typescript source file into JSON-Schema.
```
Zombiefruit: What part of the source file does it transform? Just the types? Logic?

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
...
+    import path from 'path';
+    import compilerOptions from './utils';
+
+    const projectPath = '/Projects/MySecretProject/;
+    const fileName = projectPath + 'secret-stuff.ts'
+    // TS setup
+    const program = typescript.createProgram([fileName], compilerOptions);
+    const sourceFile = program.getSourceFile(fileName);
+
+    const schema = transform(program.getTypeChecker(), sourceFile, fileName, 'MySecretProject', path.posix);
```
Zombiefruit: Maybe show the resulting schema as well?

packages/schema-extract/README.md
JoMarton: mention that it is being used by default when using createLinker. No need to use this specifically unless requiring non-linked schema

packages/schema-extract/README.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+## Linker
```
JoMarton: Consider moving this part to above ts transform, but keep custom linker in the end

packages/schema-extract/README.md
danielstro: Oh yeah, that's a mistake, there are 2 parameters

packages/schema-extract/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+The linker receives a file name and the name of an export in that file and returns the linked schema of that export.
+
+How is it any different than the `ts transformer`? When we transform a schema using the `ts transformer`, the result we get will only reference other types (Import or from the same file) using the $ref property.
+
+The linker flattens some of these types by linking them together.
+
+#### Linking rules
+In order to avoid running into infinite loops, the linker does not link every member of every schema. It does link:
```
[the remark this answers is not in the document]
danielstro: I was thinking about it but it will be a long list, and without examples I don't think it will be much clearer.

packages/schema-extract/README.md
```diff
@@ -0,0 +1,246 @@
...
+Schema extract transforms code files into [JSON-Schema](https://json-schema.org/) schemas. Schema extraction is a necessary step to create automated documentation for your code.
+
+It is composed of two parts, the **ts transformer** and **linker**.
+
+
+## Playground
+The [schema-playground](../schema-playground) is a playground that allows live editing of your code and transforms it to a schema and displays the results as you type it.
+
+## Usage
+By default, the linker is connected to our TS transformer, it uses it to transforms the code and then link it together.
```
Zombiefruit: "which uses it to transform"*

Zombiefruit: I would also specify that it's transforming the types of the code (from my understanding) and not necessarily reflecting the logic.

packages/schema-extract/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+
+## Playground
+The [schema-playground](../schema-playground) is a playground that allows live editing of your code and transforms it to a schema and displays the results as you type it.
+
+## Usage
+By default, the linker is connected to our TS transformer, it uses it to transforms the code and then link it together.  
+If you want to use the linker with a different extractor please see the [creating a custom linker section](#creating-a-custom-linker)
+
+Import the `createLinker` function from `@ui-autotools/schema-extract`. This function receives a string array of file paths and returns an initialized linker class.
```
Zombiefruit: "To use the Schema Extract project to extract type data from your files, import the `createLinker`..." I would just make it very clear here that _this_ is how you use Schema Extract and that using the transformer/linker are internal/more specific methods

packages/schema-extract/README.md
```diff
@@ -0,0 +1,244 @@
...
+Schema extract transforms all the types (Explicit or inferred) in your code files into [JSON-Schema](https://json-schema.org/) schemas. Schema extraction is a necessary step to create automated documentation for your code.
+
+It is composed of two parts, the **ts transformer** and **linker**.
+
+
+## Playground
+The [schema-playground](../schema-playground) is a playground that allows live editing of your code and transforms it to a schema and displays the results as you type it.
+
+## Usage
+By default, the linker is connected to our **TS transformer**, which uses it to transforms the code, extracts all the types in it and then link them together.
```
Zombiefruit: I don't think this line is necessary.

packages/schema-extract/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+To use the Schema Extract project to extract type data from your files, import the `createLinker` function from `@ui-autotools/schema-extract`. This function receives a string array of file paths and returns an initialized linker class.
+
+After creating a linker, invoke `linker.flatten`. It receives two arguments and will return a linked schema of the entity we requested (In `exportName`):
+|Name|Type|Description|
```
Zombiefruit: This table isn't rendering for some reason.

PR #175, opened 2018-11-07
Sanity: add event listener leak test

Zombiefruit:
This PR adds a test to the sanity suite, which checks to make sure that all event listeners added to either the window, document, or body are removed after unmounting.


packages/sanity/src/hydration-test/flood-events.tsx
```diff
@@ -0,0 +1,167 @@
+import React from 'react';
+
+export class FloodEvents extends React.Component {
```
JoMarton: Does this name make sense? I understand this component's use is to "flood" all events but it doesn't really describe the contents of the class

packages/sanity/src/hydration-test/event-listener-test.tsx
```diff
@@ -0,0 +1,62 @@
...
+import ReactDOM from 'react-dom';
+import Registry, {getCompName} from '@ui-autotools/registry';
+import chai, {expect} from 'chai';
+import sinonChai from 'sinon-chai';
+import {hydrate} from 'react-dom';
+import {FloodEvents} from './flood-events';
+import {overrideEventListeners} from './override-event-listeners';
+
+chai.use(sinonChai);
+
```
JoMarton: I'd like an overall comment here explaining the approach since this isn't trivial

packages/sanity/src/hydration-test/event-listener-test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      describe(getCompName(Comp), () => {
+        before(() => {
+          ReactDOM.render(<FloodEvents />, root);
+          ReactDOM.unmountComponentAtNode(root);
+        });
+
+        componentMetadata.simulations.forEach((simulation) => {
+          it('component should unmount without leaving event listeners on the window, document, and body', () => {
+            const {windowEe, documentEe, bodyEe} = overrideEventListeners();
+            const matchEverything = /.*/;
```
JoMarton: you can put that up top, no need to define every run

packages/sanity/src/hydration-test/override-event-listeners.ts
```diff
@@ -0,0 +1,51 @@
...
+  documentEe: EventEmitter;
+  bodyEe: EventEmitter;
+}
+
+export function overrideEventListeners(): IEventEmitterSet {
+  const windowEe = new EventEmitter();
+  const documentEe = new EventEmitter();
+  const bodyEe = new EventEmitter();
+
+  const oldWindowEventListener = window.addEventListener;
```
JoMarton: I think this file could be refactored to use a common method

packages/sanity/src/hydration-test/event-listener-test.tsx
```diff
@@ -0,0 +1,62 @@
...
+            hydrate(componentMetadata.simulationToJSX(simulation), root);
+            ReactDOM.unmountComponentAtNode(root);
+            index++;
+
+            const windowListeners = windowEe.getListeners(matchEverything);
+            const documentListeners = documentEe.getListeners(matchEverything);
+            const bodyListeners = bodyEe.getListeners(matchEverything);
+
+            Object.entries(windowListeners).forEach((listener) => {
+              const listenerType = listener[0];
```
JoMarton: all of these 3 test loops can use a single common method

packages/sanity/src/hydration-test/event-listener-test.tsx
```diff
@@ -51,23 +48,22 @@ export const eventListenerTest = (): void => {
...
 
-        afterEach(() => {
-          eventReset();
-        });
-
         componentMetadata.simulations.forEach((simulation) => {
           it('component should unmount without leaving event listeners on the window, document, and body', () => {
-            const {windowEe, documentEe, bodyEe, reset} = overrideEventListeners();
-            eventReset = reset;
+            const {windowLogger, documentLogger, bodyLogger} = overrideEventListeners();
```
alisey: I would unwrap the function and create the loggers here,
```
const windowLogger = attachEventListenerLogger(window);
const documentLogger = attachEventListenerLogger(document);
const bodyLogger = attachEventListenerLogger(document.body);
```
... two extra lines, but it would nicely mirror those 3 detach and assert calls.

PR #180, opened 2018-11-11
Accessibility component flag - skip running a11y tests

RashelleKosib:


packages/mock-repo/src/non-a11y-component/non-a11y-component.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+export class NonA11yComponent extends React.Component<{}> {
```
Zombiefruit: Should just be:
```suggestion
export class NonA11yComponent extends React.Component {
```

packages/mock-repo/src/non-a11y-component/non-a11y-component.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
```
Zombiefruit:
```suggestion
import React from 'react';
```

packages/mock-repo/src/non-a11y-component/non-a11y-component.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export class NonA11yComponent extends React.Component<{}> {
+
+  public render() {
+    return (
+      <button id="saveChanges" role="NOT AN ARIA ROLE">
```
Zombiefruit: I think the id and the text content are unnecessary. 
```suggestion
      <button role="NOT AN ARIA ROLE" />
```

packages/registry/test/registry/registry.spec.tsx
```diff
@@ -35,6 +35,10 @@ describe('Registry', () => {
     it('throws if a component name is not valid', () => {
       expect(() => Registry.getComponentMetadata(InvalidNameComp)).to.throw();
     });
+
+    it('returns default value equals to true for a11yCompliant', () => {
```
Zombiefruit: This test is in the wrong place. It should be in the `component-metadata.spec.tsx` file.

But, although testing is good (👍 ) I actually don't think we need a test for this

PR #183, opened 2018-11-18
Add Github integrations for Snap

Zombiefruit:


packages/snap/src/snap-test/set-batch-id.ts
```diff
@@ -0,0 +1,30 @@
...
+function getHeadHash() {
+  return execSync('git rev-parse --verify HEAD').toString();
+}
+
+function getParentsHashArray() {
+  const headCommitHash = getHeadHash();
+  return execSync(`git rev-list --parents -n 1 ${headCommitHash}`).toString().split(' ');
+}
+
+function getPRHeadHash() {
```
alisey: I think the name of the function is confusing, since it checks for `isPullRequest` it doesn't always return the PR head hash, the question then is what it returns. I don't want to have any magical code that we don't understand in the project.

packages/snap/src/snap-test/set-batch-id.ts
```diff
@@ -0,0 +1,30 @@
...
+  const parentHashIndex = isPullRequest ? PULL_REQUEST_PARENT_HASH_INDEX : HEAD_HASH_INDEX;
+  return parentsHashArr[parentHashIndex].trim();
+}
+
+export function setApplitoolsId() {
+  let batchId;
+  try {
+    batchId = getPRHeadHash();
+  } catch (e) {
+    batchId = process.env.BUILD_VCS_NUMBER;
```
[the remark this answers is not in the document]
Zombiefruit: It's a Teamcity variable. I'll add a comment.

https://confluence.jetbrains.com/display/TCD9/Predefined+Build+Parameters

packages/snap/src/snap-test/set-batch-id.ts
Zombiefruit: Added warning regarding the magical quality of said method

PR #184, opened 2018-11-19
MAJOR change - changed the a11y  flag pattern

RashelleKosib:


README.md
```diff
@@ -101,6 +101,13 @@ const meta = Registry.getComponentMetadata(compWithUnsafeLifecycle);
 meta.reactStrictModeCompliant = false;
 ```
 
+Components are also assumed by default to be a11y compliant. If you want to override it (meaning your component is unAccessible) - set the nonA11yCompliant flag in the metadata to true, e.g:
```
Zombiefruit: Components are also assumed by default to be a11y compliant. If you want to override it (meaning your component is unAccessible) - set the nonA11yCompliant flag in the metadata to true, e.g:
```suggestion
Components are also assumed by default to be [axe-core](https://github.com/dequelabs/axe-core) compliant. If your component is not axe-core compliant, set the `nonA11yCompliant` flag in the metadata to true, e.g:
```

RashelleKosib: Don't you Annie me

PR #187, opened 2018-11-19
Add a flag to allow opting out from the event listener check

Zombiefruit:


packages/sanity/src/hydration-test/event-listener-test.tsx
```diff
@@ -30,50 +30,56 @@ export const eventListenerTest = (): void => {
...
+              if (errors.length) {
+                throw Error(errors.join('\n'));
+              }
+            });
           });
         });
-      });
+      } else {
+        // tslint:disable-next-line:no-console
+        console.error(`Skipping event listener test for Component: "${componentName}". The "nonEventListenerCheckCompliant" flag was set to true in the metadata.`);
```
JoMarton: why `console.error`. It isn't an error if someone meant for the test not to run no?

packages/registry/test/registry/component-metadata.spec.tsx
```diff
@@ -59,9 +59,17 @@ describe('Component Metadata', () => {
     });
   });
 
-  describe('A11y metadata', () => {
-    it('returns default value equals to false for nonA11yCompliant', () => {
+  describe('Compliancy flags', () => {
```
RashelleKosib: I think the title should be more informative about the flags (only because the file name is not visible in the describe titles)

PR #191, opened 2018-11-22
[WIP] Initial POC of user config

yardnsm:
This is a small POC of a user config feature. Initially, this was intended to allow the support for JavaScript files in the `sanity` tool by setting up custom require hooks.

An example config file that allows JavaScript support with css-modules require hooks is as follows:

```js
const {wixCssModulesRequireHook} = require('yoshi-runtime');
const {attachHook} = require('@stylable/node');

module.exports = {
  setup() {

    // Setup require hook for babel
    require('yoshi/src/require-hooks');

    // Setup require hook for css modules
    wixCssModulesRequireHook('./src');

    // Setup Stylable
    attachHook({});
  },

  // Disable `autotools`' default TypeScript and Stylable require hooks
  initializeRequireHooks: false
};
```

### Why not adding these require hooks directly in `autotools`?

* Some project relies on different versions of `yoshi` and `babel`, so it'll be better if the consumers will handle this by themselves.

* This will spare the addition of more dependencies to `autotools`.

### DIFF
--- a/packages/scripts/src/cli.ts
+++ b/packages/scripts/src/cli.ts
@@ -9,15 +9,28 @@ import {a11yTest, impactLevels} from '@ui-autotools/a11y';
 import {buildWebsite, startWebsite} from '@ui-autotools/showcase';
 import ssrTest from './ssr-test/mocha-wrapper';
 import {importMetaFiles} from './import-meta-files';
-import {registerRequireHooks} from '@ui-autotools/utils';
+import {getUserConfig, registerRequireHooks} from '@ui-autotools/utils';
 
 dotenv.config();
-registerRequireHooks();
 
 const program = new Command();
 const projectPath = process.cwd();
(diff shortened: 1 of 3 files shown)


alisey:
The idea of adding a config file is reasonable, but there's one issue with defining require hooks in the same file as other settings (of which we have none at the moment) - this would prevent us from bundling the config for the browser environment. Not sure how much of an issue this is, but I'd prefer if the setup function was defined in a separate file:

```
// .autotools/node-require-hooks.js

module.exports = () => {
  require('yoshi/src/require-hooks');
  require('yoshi-runtime'). wixCssModulesRequireHook('./src');
  require('@stylable/node').attachHook();
};
```

If this file exists then autotools would use, and otherwise it would its default set of hooks. What do you think?

yardnsm:
@alisey Sounds good to me.

alisey:
Implemented in #203

PR #200, opened 2018-11-29
CLI - created cli scripts for sanity / snap / a11y

danielstro:
## Describe this Pull Request
Currently all scripts are run by using the @ui-autotools/scripts packages. This PR adds scripts to the sanity, snap & a11y packages so that they can be run separately if needed.

### Types of changes
- [ ] None of the above

- [ ] Patch
- [x] Minor
- [ ] Major

- [ ] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ npm i --save-dev @ui-autotools/snap
 ```
 then run the following command:
 ```shell
-autotools snap --files ./components/**/*.meta.ts
+autotools-snap --files ./components/**/*.meta.ts
 ```
 
 #### Options
--- a/README.md
+++ b/README.md
@@ -117,9 +117,6 @@ meta.nonEventListenerTestCompliant = true;
 ```
 
 ## CLI Tools
-We have created a package that contains all of the commands to run our tools. If you only want to use some of our packages, you can install each package you want and it comes with its own command.
-
-**Note**: You cannot use `showcase` separately from the other packages at the moment. You will have to install the scripts packages to run it.
 ### Universal Options
 
 - `files`: glob pattern used to match metadata files. Defaults to `src/**/*.meta.ts?(x)`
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -117,7 +117,9 @@ meta.nonEventListenerTestCompliant = true;
 ```
 
 ## CLI Tools
+We have created a package that contains all of the commands to run our tools. If you only want to use some of our packages, you can install each package you want and it comes with its own command.
```
JoMarton: I think we want to direct people to consume the packages separately and deprecate the `scripts` repo soon. So we don't have to put in this paragraph

packages/utils/src/cli-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+import { registerRequireHooks } from './';
+import path from 'path';
+
+export function cliInit(): void {
+    dotenv.config();
+    registerRequireHooks();
+}
+
+export const defaultMetaGlob = 'src/**/*.meta.ts?(x)';
+export function getDefaultWebpackConfigPath(projectPath: string): string {
```
alisey: I'd rename this to `getProjectWebpackConfigPath()`, or simply export `'.autotools/webpack.config.js'`.

README.md
```diff
@@ -139,27 +138,39 @@ Sanity uses puppeteer to test client-side hydration. Results are printed in the
...
 
 #### Usage
-
+First install the package 
+```shell
+npm i --save-dev @ui-autotools/snap
+```
+then run the following command:
 ```shell
 autotools snap --files ./components/**/*.meta.ts
```
JoMarton: `autotools-snap`

packages/sanity/src/cli.ts
```diff
@@ -0,0 +1,23 @@
...
+const program = new Command();
+const projectPath = process.cwd();
+const webpackConfigPath = getWebpackConfigPath(projectPath);
+
+program
+    .description('run sanity checks on all components with a metadata description')
+    .option('-f, --files [pattern]', 'Grep file')
+    .action((options) => {
+        const metaGlob: string = options.files || defaultMetaGlob;
+        glob.sync(metaGlob, { absolute: true, cwd: projectPath }).forEach(require);
```
alisey: We have an `importMetaFiles()` function in the `scripts` package, use it here or add a TODO comment.

PR #203, opened 2018-12-03
CLI - Adding user config file

danielstro:
## Describe this Pull Request
Adding user config file. If the the file is present and in the .autotools directory it will be required, if not we will use our default hooks.

### Types of changes
- [ ] None of the above

- [ ] Patch
- [x] Minor
- [ ] Major

- [x] Not sure.
- [ ] No


packages/utils/src/require-hooks/require-hooks.ts
```diff
@@ -1,4 +1,13 @@
+import fs from 'fs';
+import path from 'path';
+
 export function registerRequireHooks() {
-  require('@ts-tools/node');
-  require('@stylable/node/register');
+  const projectPath = process.cwd();
```
danielstro: We are also using the projectPath in the CLI commands. Should we receive as a prop to avoid duplication?

JoMarton: I think so. 
Regarding documnetation it - we definitly need to document it! we're adding a way to configure our tools so we must write it down in the README

README.md
```diff
@@ -191,4 +191,10 @@ autotools showcase --files src/**/*.meta.ts
 To build a static website:
 ```shell
 autotools showcase --files src/**/*.meta.ts --output build/website
-```
\ No newline at end of file
+```
+
+## User config
+`@ui-autotools` assumes that your code is using [Typescript](http://typescriptlang.org) or [Stylable](https://stylable.io/), and therefor automatically requires hooks to hanlde such files (`.ts`, `.st.css` etc.). If however you are not using these or you are using different hooks, we support requiring your own hooks.
```
alisey:
```suggestion
`@ui-autotools` assumes that your code is using [Typescript](http://typescriptlang.org) and [Stylable](https://stylable.io/), and therefor automatically requires hooks to handle such files (`.ts`, `.st.css` etc.). If however you are not using these or you are using different hooks, we support requiring your own hooks.  
```

README.md
```diff
@@ -191,4 +191,10 @@ autotools showcase --files src/**/*.meta.ts
...
 autotools showcase --files src/**/*.meta.ts --output build/website
-```
\ No newline at end of file
+```
+
+## User config
+`@ui-autotools` assumes that your code is using [Typescript](http://typescriptlang.org) or [Stylable](https://stylable.io/), and therefor automatically requires hooks to hanlde such files (`.ts`, `.st.css` etc.). If however you are not using these or you are using different hooks, we support requiring your own hooks.  
+In the root of your project, inside the `.autotools` folder, create a new file named `node-require-hooks.js`, require your hooks and invoke them and `@ui-autotools` will use your config file.  
+
+**Note:** If you are using your own config file, keep in mind that `@ui-autotools` will **not use its default config file**.
```
alisey:
```suggestion
**Note:** If you are using your own config file, keep in mind that `@ui-autotools` will **not use its default set of hooks**.
```

packages/utils/src/index.ts
```diff
@@ -5,7 +5,7 @@ export * from './console';
 export * from './get-project-name';
 export * from './http';
 export * from './raw-asset-webpack-plugin';
-export {registerRequireHooks} from './require-hooks/require-hooks';
+// export {registerRequireHooks} from './require-hooks/require-hooks';
```
alisey:
```suggestion
```

packages/utils/src/index.ts
```diff
@@ -5,7 +5,7 @@ export * from './console';
...
 export {compilerOptions} from './typescript/constants';
 export {createTsService} from './typescript/createMemoryTsService';
-export * from './cli-utils';
+// export * from './cli-utils';
```
alisey:
```suggestion
```

PR #221, opened 2018-12-19
schema-views: add support for view variants, fix runtime errors

alisey:
* Add support for view variants (summary, expanded, etc)
* Add expanded view for interfaces
* Fix runtime errors related to `.properties` access in schemas without that key

### Types of changes
- [ ] None of the above

- [x] Patch
- [ ] Minor
- [ ] Major

- [x] No

### DIFF
--- a/packages/schema-views/src/schema-view-registry.ts
+++ b/packages/schema-views/src/schema-view-registry.ts
@@ -2,18 +2,34 @@ import {Schema} from './schema';
 import {SchemaView} from './schema-view';
 import {SchemaPredicate} from './schema-predicates';
 
-export class SchemaViewRegistry {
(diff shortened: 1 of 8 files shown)


packages/schema-views/src/schema-view-registry.ts
```diff
@@ -2,18 +2,34 @@ import {Schema} from './schema';
 import {SchemaView} from './schema-view';
 import {SchemaPredicate} from './schema-predicates';
 
+interface IViewRegistration {
+  view: SchemaView;
+  predicate: SchemaPredicate;
+  variant?: string;
+}
+
```
danieleww: We could use a more specific type for the variant, like:

type SchemaViewVariant = 'expanded' | 'summary'

...
variant?: SchemaViewVariant
...

Issue #223, opened 2018-12-19
Function arguments without type declaration is treated as non-required argument


danieleww:
Function arguments without type declaration is treated as non-required argument

```
export function foo(a) {
}
```

Schema:
```
{
      "$ref": "common/function",
      "arguments": [
        {
          "name": "a"
        }
      ],
      "returns": {
        "$ref": "common/undefined"
      }
}
```

Expected to have:
```
"requiredArguments": [
    "a"
]
```

Issue #219, opened 2018-12-18
Argument names of fat-arrow-functions get lost in schema-extract


danieleww:
Argument names of fat-arrow-functions get lost in schema-extract

```
const foo = (x:string) => x
```

```
"foo": {
      "$ref": "common/function",
      "returns": {
        "schema": {
          "type": "string"
        }
      },
      "arguments": [
        {
          "schema": {
            "type": "string"
          }
        }
      ]
    },
```

PR #224, opened 2018-12-19
Daniele/inline function view

danieleww:
### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] None of the above

- [ ] Patch
- [ ] Minor
- [ ] Major

- [x] No

### DIFF
--- a/packages/schema-views/src/views/types/function.tsx
+++ b/packages/schema-views/src/views/types/function.tsx
@@ -13,7 +13,7 @@ export const FunctionTypeView: React.FunctionComponent<ISchemaViewProps> = (prop
     const optional = required.includes(arg.name) ? '' : '?';
     const argName = arg.name + optional;
     return [
-      argName + ':',
+      argName + ': ',
       (
         <BaseView
           schemaRegistry={props.schemaRegistry}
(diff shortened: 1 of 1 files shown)


danieleww:
```
export function foo1 (x: ((a:number)=>number) | string = "", ...args:string[]){
	return true;
}


export function foo2 (x: string | ((a:number)=>number) = "", ...args:string[]){
	return true;
}


export function foo3 (x: string | ((a:number)=>number) = (u)=>u, ...args:string[]) {
	return true;
}


export function foo4 (x: {x:number, y:string} = {x:1, y:""}, ...args:string[]) {
	return true;
}
```

danieleww:
issues found:
- https://github.com/wix-incubator/ui-autotools/issues/223
- https://github.com/wix-incubator/ui-autotools/issues/219

packages/schema-views/src/views/types/function.tsx
```diff
@@ -8,34 +8,25 @@ import style from './type.st.css';
...
+      argName + ': ',
       (
         <BaseView
           schemaRegistry={props.schemaRegistry}
           viewRegistry={props.viewRegistry}
           schema={arg}
         />
       ),
-      defaultValue ? ` = ${defaultValue}` : ''
+      arg.hasOwnProperty('default') ? ` = ${JSON.stringify(arg.default)}` : (
```
alisey: I think `'default' in args` would express the intent more clearly. I was confused for a moment thinking about why would we want to exclude properties coming from the prototype, to only realize that the actual intent is to correctly handle default when it's falsy, something that I missed when I initially wrote this function.

PR #226, opened 2018-12-27
Sanity - added flag for hydration tests

danielstro:
## Describe this Pull Request
Added a flag to the metafiles that allows disabling hydration tests

### Types of changes
- [ ] None of the above

- [ ] Patch
- [x] Minor
- [ ] Major

- [x] Yes and I updated README.md
- [ ] No

### DIFF
--- a/packages/registry/src/registry/component-metadata.tsx
+++ b/packages/registry/src/registry/component-metadata.tsx
@@ -8,6 +8,7 @@ export class ComponentMetadata<Props, State> implements IComponentMetadata<Props
   public nonReactStrictModeCompliant: boolean = false;
   public nonA11yCompliant: boolean = false;
   public nonEventListenerTestCompliant: boolean = false;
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -116,6 +116,13 @@ const meta = Registry.getComponentMetadata(nonEventListenerTestCompliant);
 meta.nonEventListenerTestCompliant = true;
 ```
 
+In some cases you may want to cancel our hydration tests (React is escaping style attributes and that may cause the test to fail). If you wish to skip this test, set the `nonHydrationTestCompliant` flag in the metadata to true.
+
+```ts
+const meta = Registry.getComponentMetadata(nonHydrationTestCompliant);
```
JoMarton: I think this line could be confusing, people might think they need this line to disable the test. I know all other flags used this pattern, so please remove this line from them as well

PR #235, opened 2019-01-22
snap - run snapshots on bundled files poc

danielstro:
## Describe this Pull Request
POC to run the snap tool on files after building

### Types of changes
- [ ] None of the above

- [ ] Patch
- [x] Minor
- [ ] Major

- [x] No

### DIFF
--- a/packages/registry/src/registry/types.ts
+++ b/packages/registry/src/registry/types.ts
@@ -6,6 +6,11 @@ export interface IRegistry<AssetMap = any> {
   clear: () => void;
 }
 
+export interface ICompiledComponent {
+  cssPath?: string;
+  compPath: string;
+}
+
 export interface IComponentMetadata<Props, State> {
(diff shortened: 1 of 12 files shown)


packages/registry/src/registry/types.ts
```diff
@@ -6,6 +6,11 @@ export interface IRegistry<AssetMap = any> {
   clear: () => void;
 }
 
+export interface ICompiledComponent {
+  cssPath?: string;
```
Zombiefruit: We're renaming these right?

danielstro: Since I removed it to be under the compiledComponent attribute, I don't think we need to rename it. It is more obvious that this is the css path for the compiled product:
```metaData.compiledComponent.cssPath```

Zombiefruit: Oh yeah, cool 💯 🥇

PR #241, opened 2019-02-19
registry: Add getCompId

hepiyellow:
Add API for `getCompId` which ensures a unique id ,
as opposed to `getCompName`, which in the future may not be unique,
and also has other name restrictions (like being only `[a-zA-Z\.]*`)


packages/registry/src/registry/registry.ts
```diff
@@ -3,14 +3,16 @@ import {ComponentType} from 'react';
 import {ComponentMetadata} from './component-metadata';
 import Metadata from './metadata';
 import {getCompName} from '../utils/get-comp-name';
+import {getCompId} from '../utils/get-comp-id';
```
alisey: Let's just have `.getId()` on the component metadata.

packages/registry/src/registry/registry.ts
```diff
@@ -3,14 +3,16 @@ import {ComponentType} from 'react';
...
 import {isValidComponentName} from '../utils';
 
 const Registry: IRegistry = {
   metadata: new Metadata(),
   getComponentMetadata <Props, State>(comp: ComponentType<Props>): ComponentMetadata<Props, State> {
     const newCompName = getCompName(comp);
+    const newCompId = getCompId(comp);
 
-    if (!newCompName) {
+    if (!newCompName || !newCompId) {
```
alisey: We don't need to check this here. This is for testing the presence of name of displayName, not ID. Maybe add another test for ID with a different error message.

packages/registry/test/fixtures/component-fixtures.tsx
```diff
@@ -9,19 +9,23 @@ export const TestComp: React.FunctionComponent<ITestProps> = (props: ITestProps)
 };
 
 TestComp.displayName = 'TestComp';
+Object.freeze(TestComp);
```
alisey: Why do you need to freeze them?

Zombiefruit:
Hey, I'm not aware of the background for this PR @hepiyellow, but I'm just wondering why compName isn't enough? We already ensure that every component's name is unique and, from my knowledge at least, I don't think we're planning on removing that restriction.

packages/registry/src/registry/registry.ts
hepiyellow: I want to add a `getId()` on the `ComponentMetadata`, but we need a `getCompId` outside of it, so we can check for duplicate ids BEFORE we even create a ComponentMetadata.

packages/registry/src/registry/registry.ts
hepiyellow: ok

packages/registry/test/fixtures/component-fixtures.tsx
hepiyellow: Becuase I saw that tests modify the display name:
```
TestComp.displayName = "ReturnDisplayName".
```

This means that one test affects other tests.

hepiyellow:
@alisey rereview please

hepiyellow:
@Zombiefruit As I wrote before:
>Add API for `getCompId` which ensures a unique id ,
as opposed to `getCompName`, which in the future may not be unique,
and also has other name restrictions (like being only `[a-zA-Z\.]*`)

My first choice would be to rename `getCompName` -> `getCompId`.
I guess we don't want to make a breaking change.
So adding `getCompId` would not break it.
By adding `getCompId` we are freeing `getCompName` from having to be unique in the Future.

BTW:
I am also fine if we add JSDoc on `getCompName` and say that it is unique. That is, if you "promise" me (sonusmers) that it is unique (not only in implementation, but in API).

Zombiefruit:
@hepiyellow 
We definitely don't have any plans to change the uniqueness of `compName`. They should always be unique. 

Maybe we can find a better name for `compName`, so that it's more clear that each name is unique. As far as adding `getCompId`, I'm just reluctant to add code duplication to the project, as currently it's doing exactly what `getCompName` is doing, just under a different name. 

I'm totally in favour of adding better docs, and making it clear that `compName` is, and will be, unique.

packages/registry/test/fixtures/component-fixtures.tsx
alisey: Ah, ok. But tests shouldn't modify displayName, that's a mistake.

packages/registry/src/registry/registry.ts
alisey: We should check for duplicate IDs not __before__, but __when__ someone is registering an ID for the component. Since ID is the component's name, and we already have validation for that, you don't need to do anything.

You can simply add `.getId()` on the component metadata and that's it, it's going to work as expected and throw an error if you have components with duplicate names.

PR #267, opened 2020-04-11
feat: upgrade to stylable@3

AviVahl:
## Describe this Pull Request
Upgrade repository to use `stylable@3`.

### Types of changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] None of the above

- [ ] Patch
- [ ] Minor
- [x] Major

- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
     "prettify": "npx prettier \"./**/*.{js,ts,tsx}\" --write"
   },
   "devDependencies": {
-    "@stylable/cli": "^1.4.0",
-    "@stylable/core": "^1.4.0",
+    "@stylable/cli": "^3.4.1",
+    "@stylable/core": "^3.4.1",
     "@ts-tools/node": "^1.1.3",
     "@ts-tools/webpack-loader": "^1.1.4",
     "@types/chai": "^4.2.11",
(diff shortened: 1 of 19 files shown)


AviVahl:
While this passes, there are several places in the code (e.g. `IComponentMetadata['styles']`) that use `Map<any, IStyleMetadata>`, where the `any` is actually `RuntimeStylesheet` **instance**, and the `IStyleMetadata` contains information relating to the build (aka wtf).

Not sure whether this is a leftover from `snap` package and its need to know about assets for snapshotting the components.

Perhaps @Zombiefruit or @alisey could clarify?

I currently left those imports as `import style1 from '...'`, and it looks like nothing in the flow is really using the key of that map.

Zombiefruit:
Hey @AviVahl, yeah, the component metadata styles were only used in snap and showcase, where (in snap's case) we'd create a snapshot per variant. I honestly can't remember exactly why we solved it this specific way. There was lots of discussion about this. Various fragments from my memory include something about the Stylable compiler not being able to resolve imports, and allowing users to specify however many static resources they'd want (including stuff that affected things globally).

